PR #7, opened 2025-09-14
[Tech Debt] Update gitignore

ohadedry:


.gitignore
```diff
@@ -1,3 +1,7 @@
+# AM specific
+demo
```
aliabufoul: I don't see demo

.gitignore
```diff
@@ -1,3 +1,7 @@
...
+tests/test_commands/data/*
```
aliabufoul, line 2: a question, are there no cases that we might need to modify the data files?
I undsteand that these files might get updated by tests executions and we don't want to commit thos changes

ohadedry: Data files are not supposed to be changed

PR #13, opened 2025-09-17
fix: Change output font color to white in json format

aviatco:
Issue: When using the JSON output format, the result is incorrectly displayed in grey instead of white.
Solution: This PR updates the print_output_format function to use _safe_print for JSON output, replacing print_grey to avoid applying the grey font color.


ohadedry (approved):

ayeshurun:
@aviatco please add PR description :)

PR #19, opened 2025-09-21
fix(folders): newly created item is not displaying when running ls in a folder path 

aviatco:
Issue: When FAB_FOLDER_LISTING_ENABLED is set to true, creating a new item under a folder and then running ls does not show the newly created item in the listing.
Root Cause: The folder cache is not being invalidated—only the item cache is refreshed.
Fix: We need to ensure the folder cache is also invalidated to reflect the latest changes in the listing.

### DIFF
--- a/src/fabric_cli/utils/fab_mem_store.py
+++ b/src/fabric_cli/utils/fab_mem_store.py
@@ -438,8 +438,14 @@ def get_item_id(workspace: Workspace, name) -> str:
 
 
 def upsert_item_to_cache(item: Item) -> None:
-    # Due to dependent elements (e.g. SQLEndpoint for Lakehouse), we need to invalidate the cache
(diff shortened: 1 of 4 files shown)


src/fabric_cli/utils/fab_mem_store.py
```diff
@@ -438,8 +438,12 @@ def get_item_id(workspace: Workspace, name) -> str:
 
 
 def upsert_item_to_cache(item: Item) -> None:
-    # Due to dependent elements (e.g. SQLEndpoint for Lakehouse), we need to invalidate the cache
+    # Invalidate both item cache and folder cache to maintain consistency
```
ayeshurun, line 441: Please add unit tests for ensure the fix is working as expected

PR #27, opened 2025-10-05
feat: support dataflow CRUD

aviatco:
# 📥 Pull Request

Support Dataflow item type

* Enable CRUD commands for dataflow. 
* Update docs

**Supported commands:**
  * cd 
  * ls
  * create
  * get
  * set
  * rm
  * import 
  * export 
  * cp 
  * mv
  * open

### DIFF
--- a/src/fabric_cli/core/fab_types.py
+++ b/src/fabric_cli/core/fab_types.py
@@ -554,7 +554,7 @@ uri_mapping = {
     ItemType.COPYJOB: "copyjobs",
     ItemType.GRAPHQLAPI: "graphql",
     ItemType.MOUNTED_DATA_FACTORY: "mounteddatafactories",
-    ItemType.DATAFLOW: "dataflows",
+    ItemType.DATAFLOW: "dataflows-gen2",
 }
 
 # Item Payload definition


src/fabric_cli/core/fab_config/command_support.yaml
```diff
@@ -166,6 +166,7 @@ commands:
           - variable_library
           - graph_ql_api
           #- lakehouse # this is included just to support underlying /Files
+          - dataflow
```
ayeshurun, line 169: @aviatco please add to export command as well

ayeshurun:
@aviatco can you please elaborate what are the supported commands that enables CRUD? (open, cd, rm, set, get, mkdir ?).

src/fabric_cli/core/fab_types.py
```diff
@@ -552,6 +554,7 @@ class MirroredDatabaseFolders(Enum):
     ItemType.COPYJOB: "copyjobs",
     ItemType.GRAPHQLAPI: "graphql",
     ItemType.MOUNTED_DATA_FACTORY: "mounteddatafactories",
+    ItemType.DATAFLOW: "dataflows",
```
ayeshurun: have you validated the open command? it seems it should be `dataflows-gen2`but perhaps i am wrong

src/fabric_cli/core/fab_config/command_support.yaml
aviatco: it is in the export - see below

src/fabric_cli/core/fab_config/command_support.yaml
```diff
@@ -239,6 +241,7 @@ commands:
           - copy_job
           - variable_library
           - graph_ql_api
+          - dataflow
```
aviatco, line 244: @ayeshurun here is the export

PR #29, opened 2025-10-15
fix: 2 in-support versions of Python

dzsquared:
# 📥 Pull Request

The GA version of Fabric CLI doesn't support 3.13 or 3.14, but previews published to pypi do. When users on a more recent version of python pip install fab the preview version silently installs and then they're left wondering why the behavior is wildly different.

https://devguide.python.org/versions/

Previous discussions about this issue have been lost in repo and channel changes, but I believe there were dependency issues preventing this update previously. I haven't encountered issues with the update, but would like more experienced input.

### DIFF
--- a/.github/workflows/fab-build.yml
+++ b/.github/workflows/fab-build.yml
@@ -79,6 +79,10 @@ jobs:
             tox-env: "py311"
           - python-version: "3.12"
             tox-env: "py312"
+          - python-version: "3.13"
+            tox-env: "py313"
+          - python-version: "3.14"
+            tox-env: "py314"
     steps:
       - uses: actions/checkout@v3
 


ayeshurun:
Hi @dzsquared ! Thank you for your interest in contributing!

This pull request is being closed as it does not reference a "help wanted" issue.
Please follow our [contribution process](https://github.com/microsoft/fabric-cli/blob/main/CONTRIBUTING.md):

1. Find or create an issue labeled "help wanted"
2. Get approval from maintainers
3. Submit your PR with a reference to the approved issue

ayeshurun:
@dzsquared we encountered a compatibility issue with Python versions 3.13 and above due to our dependency on msal[broker], which currently does not support versions beyond 3.12. As a result, we are unable to proceed with the upgrade at this time.

PR #61, opened 2025-11-10
fix(core): correct context persistence in virtual environments

### LINKED ISSUE #45: [BUG] `cd` is not working
### Bug Description

`cd` is not working

### fabric-cli version

1.2.0

### Python version

3.11.9

### Operating System

Windows

### CLI Mode

Interactive mode

### Authentication Method

Interactive browser login

### Steps to Reproduce

1. `fab auth login`
2. `fab ls` to list workspace
3. pick a workspace and `cd`
4. `fab ls`

### Expected Behavior

I would expect it to show artifacts and folders.

### Actual Behavior

Repro step 4) still list workspaces. 

### Additional Context

_No response_

### Possible Solution

_No response_

ayeshurun:
# 📥 Pull Request

This pull request fixes a bug where context persistence would fail to save the session ID when the Fabric CLI was run from within a virtual environment (venv).

- Fixes #45 

### 🐞 The Problem

The session ID, which is used to manage context files, was being resolved incorrectly inside a virtual environment. The logic did not properly traverse the process tree to find the true parent shell process, leading to a new, incorrect context being created with each command instead of persisting the existing one.

### ✅ The Solution

The session ID resolution logic in `fab_context.py` has been fixed to ensure it correctly identifies the parent shell process, even when running inside a venv.

-   **Correct Process Traversal**: The logic now correctly traverses one additional level up the process tree when a virtual environment is detected, successfully anchoring the session to the parent shell.
-   **Robust Fallback Mechanism**: The implementation uses a clean and safe fallback hierarchy to handle unexpected process states, ensuring the CLI remains stable.
-   **Code Refactoring**: The session resolution logic was centralized into a `_get_session_process` method, making the code cleaner, more maintainable, and easier to test.

#### 📝 Note on Virtual Environment Detection

The check `sys.prefix != sys.base_prefix` is used to determine if the CLI is running inside a virtual environment. This is the standard and recommended method as documented by Python's official `venv` documentation.

-   `sys.base_prefix`: Always points to the base Python installation directory.
-   `sys.prefix`: Points to the active environment's directory.

When running in a venv, `sys.prefix` will point to the venv's directory, causing the two values to differ. You can find more details in the [official Python documentation on `venv`](https://docs.python.org/3/library/venv.html).

##### [How venvs work](https://docs.python.org/3/library/venv.html#how-venvs-work)

_When a Python interpreter is running from a virtual environment, [sys.prefix](https://docs.python.org/3/library/sys.html#sys.prefix) and [sys.exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix) point to the directories of the virtual environment, whereas [sys.base_prefix](https://docs.python.org/3/library/sys.html#sys.base_prefix) and [sys.base_exec_prefix](https://docs.python.org/3/library/sys.html#sys.base_exec_prefix) point to those of the base Python used to create the environment. It is sufficient to check sys.prefix != sys.base_prefix to determine if the current interpreter is running from a virtual environment._

### DIFF
--- a/.changes/unreleased/fixed-20251111-053940.yaml
+++ b/.changes/unreleased/fixed-20251111-053940.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Fix context persistence in virtual environment
+time: 2025-11-11T05:39:40.62257109Z


src/fabric_cli/core/fab_context.py
```diff
@@ -247,6 +248,43 @@ def _process_exists(self, pid):
             # If we can't determine process existence, assume it exists to be safe
             return True
 
+    def _is_in_venv(self):
+        """Check if running inside a virtual environment."""
+        return sys.prefix != sys.base_prefix
+
+    def _get_session_process(self, parent_process: psutil.Process) -> psutil.Process:
```
may-hartov, line 255: This method currently uses parent_process, session_process, and session_parent, but the naming feels confusing. Would it make sense to rename them to something like parent and grandparent? I’d like the names to clearly indicate which process each variable refers to.

src/fabric_cli/core/fab_context.py
```diff
@@ -255,35 +293,33 @@ def _get_context_session_id(self):
...
         try:
             parent_process = psutil.Process().parent()
-            if parent_process is None:
+            if not parent_process:
                 fab_logger.log_debug(
-                    "No parent process was found. Falling back to the current process for session ID resolution."
+                    "No parent process found. Falling back to current process for session ID."
                 )
                 return os.getpid()
-        except Exception as e:
```
may-hartov, line 270: why did you remove it? what if psutil.Process().parent() throws?

ayeshurun: i merged into single try-catch block. it the snipper you mentioned throws it will still be caught inside the catch block, and the parent_process will be None, so it will return os.getpid()

src/fabric_cli/core/fab_context.py
ayeshurun: Fixed

may-hartov (approved):

PR #64, opened 2025-11-10
chore: enhance bug triage

ayeshurun:
# 📥 Pull Request

This pull request introduces several enhancements to our repository automation to streamline issue management and triage processes.

### 🐛 New Bug Triage Workflow
-   **What:** A new `/bug` command has been introduced for maintainers. When a maintainer comments `/bug` on an issue, the `needs triage` label is removed, and the `bug` label is added.
-   **Why:** This change makes the bug identification process more explicit and controlled.
-   **Related Change:** To support this new workflow, the `bug` label is no longer automatically added to issues created from the bug report template. This prevents premature labeling before a maintainer has triaged the issue.

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -199,15 +199,3 @@ configuration:
           - addReply:
               reply: >
                 This issue has been triaged and identified as a **bug**. Our team will review and prioritize it accordingly.
-
-      - description: Assign enhancement issues to HasanAboShally
-        if:
-          - payloadType: Issues
-          - isAction:
-              action: Opened
-          - hasLabel:
-              label: enhancement
-        then:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/1-bug.yml
```diff
@@ -1,7 +1,7 @@
 name: "🐛 Bug report"
 description: Create a report to help us improve fabric-cli
 title: "[BUG] "
-labels: ["bug"]
+labels: []
```
aviatco, line 4: why did you remove it?

ayeshurun: we would like to set label as bug only if triage completed and we identified it as a bug.

PR #67, opened 2025-11-13
feat(set item): support setting item definition paths

may-hartov:

- **Summary**: 

Expand and generalize the Set Item command to support setting any settable property in the item metadata and definition.

- **Context**: 
Currently, the Set Item command supports only a narrow range of properties (refer to [ITMutablePropMap](https://github.com/microsoft/fabric-cli/blob/main/src/fabric_cli/core/fab_types.py) for the supported properties).
To enhance functionality, we should explore our options to expand and generalize the Set Item command to support setting any settable property in the item metadata and definition.

Notes:
-Item definition is a complex structure that can contain hundreds of properties.
-Item metadata settable properties are displayName and description. VariableLibrary also lets you set "properties".

## Approach

1. **Updated `set item` command `-query` flag validation. Supported values:**
   - **Item metadata:**
     - `displayName`, `description`, `properties`, and `properties.`
   - **Item definition:**
     - `definition` or starts with `definition.`

2. **Backwards compatibility:**
   - If the query argument value is in `ITMutablePropMap`, map to the corresponding path.

4. **Limitations:**
   - Command doesn’t support JMESPath filters or wildcards, only explicit string paths.
   - Support a single `-query` argument value.
   - Strigified json input is not supported (for example "{\"key\":\"value\"}") - we will always try to parse json into an object
   - As `set item definition` uses `export item` to get the item definition and update the target property within it, and some properties are not included in the exported item (e.g., some default values), we support only setting properties that exist in the `get item` result.

---
## Additional Enhancements and Bug Fixes

- **Performance improvement** – Call `GET / update definition` only if the query has the `definition` prefix. Otherwise, call `GET / POST item` APIs.
- **Use specific workload URL** to get the item’s metadata – for example, use `notebooks/notebookId` rather than `items/notebookId`.
- **Change PATCH item request payload** to include only updated property rather the entire item. This ensures the API will fail if the property is readonly, and will not try to update existing, unchanged properties.
  - `get item` core API doesn’t always return the workload’s specific metadata.

### DIFF
--- a/src/fabric_cli/client/fab_api_client.py
+++ b/src/fabric_cli/client/fab_api_client.py
@@ -23,13 +23,11 @@ from fabric_cli.core.fab_exceptions import (
 from fabric_cli.errors import ErrorMessages
 from fabric_cli.utils import fab_error_parser as utils_errors
 from fabric_cli.utils import fab_files as files_utils
-from fabric_cli.utils.fab_http_polling_utils import get_polling_interval
 from fabric_cli.utils import fab_ui as utils_ui
+from fabric_cli.utils.fab_http_polling_utils import get_polling_interval
 
 GUID_PATTERN = r"([a-f0-9\-]{36})"
 FABRIC_WORKSPACE_URI_PATTERN = rf"workspaces/{GUID_PATTERN}"
-
-
 
 
 def do_request(
@@ -91,10 +89,12 @@ def do_request(
 
     # Get token
     from fabric_cli.core.fab_auth import FabAuth
+
     token = FabAuth().get_access_token(scope)
 
     # Build headers
     from fabric_cli.core.fab_context import Context as FabContext
+
     ctxt_cmd = FabContext().command
     headers = {
         "Authorization": "Bearer " + str(token),
@@ -355,10 +355,10 @@ def _poll_operation(
     args.method = "get"
     args.wait = False
     args.params = {}
-    
+
     initial_interval = get_polling_interval(original_response.headers)
     time.sleep(initial_interval)
-    
+
     while True:
         response = do_request(args, hostname=hostname)
 
@@ -373,9 +373,14 @@ def _poll_operation(
                         original_response.status_code = 200
                         return original_response
                     elif scope == fab_constant.SCOPE_FABRIC_DEFAULT:
-                        return _fetch_operation_result(
-                            args, uri, response, original_response
-                        )
+                        location_header = response.headers.get("Location", "")
+                        if location_header:
+                            return _fetch_operation_result(
+                                args, uri, response, original_response
+                            )
+
+                        original_response.status_code = 200
+                        return original_response
                 elif status == "Failed":
                     fab_logger.log_progress(status)
                     raise FabricCLIError(
--- a/.changes/unreleased/fixed-20251119-095159.yaml
+++ b/.changes/unreleased/fixed-20251119-095159.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: LRO polling - call operations/id/result API only if the location header is present
+time: 2025-11-19T09:51:59.752824783Z


docs/commands/fs/set.md
```diff
@@ -9,18 +9,47 @@ Set a workspace or item property.
...
+- `description` - The description of the item (all item types)
+- `properties` - Custom properties (`.VariableLibrary` only)
+
+### 2. Item Definition
+
+Any explicit path (specified via the `-q` / `--query` command argument) to properties within the item's `definition` structure according to [Microsoft Fabric item definitions](https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions).
+
+### Limitations
+
+- Only one property path can be specified per `-query` argument
```
ayeshurun: typo

```suggestion
- Only one property path can be specified per `--query` argument
```

ayeshurun: can you give an example?

docs/commands/fs/set.md
```diff
@@ -9,18 +9,47 @@ Set a workspace or item property.
...
+- Paths must map directly to JSON paths without filters or wildcards
```
ayeshurun:
```suggestion
- Paths must map directly to JSON paths **without** filters or wildcards
```

docs/commands/fs/set.md
```diff
@@ -9,18 +9,47 @@ Set a workspace or item property.
...
+- Only paths already present in the item definition can be updated. Properties with default values may not appear when retrieved via `get` unless explicitly set previously.
+
+### Common Item-Specific Definition Property Paths
+
+These are friendly names that map to specific paths in the item's definition structure. When using the `set` command, you can use these names directly (as the `-query` / `--q` argument value) as they map to the correct definition paths:
```
ayeshurun: i think it worthing noting here about the possibility of deprecating those friendly names sometime

src/fabric_cli/client/fab_api_client.py
```diff
@@ -373,9 +373,14 @@ def _poll_operation(
                         original_response.status_code = 200
                         return original_response
                     elif scope == fab_constant.SCOPE_FABRIC_DEFAULT:
-                        return _fetch_operation_result(
-                            args, uri, response, original_response
-                        )
+                        location_header = response.headers.get("Location", "")
```
ayeshurun, line 376: can you please explain this change

src/fabric_cli/client/fab_api_item.py
```diff
@@ -100,9 +98,24 @@ def get_item(
     return fabric_api.do_request(args)
 
 
-def update_item_definition(args: Namespace, payload: str) -> ApiResponse:
+def get_item_definition(args: Namespace) -> ApiResponse:
```
ayeshurun, line 101: nit: private method should start with `_`

e.g.,

`_get_item_defintion`

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -16,49 +17,72 @@
 def exec(item: Item, args: Namespace) -> None:
     force = args.force
     query = args.query
+    raw_string = getattr(args, "raw_string", False)
```
ayeshurun: can't we access it natively like `args.raw_string`

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -16,49 +17,72 @@
...
 
-    utils_set.validate_expression(query, item.get_mutable_properties())
+    query_value = item.get_mutable_prop_path(query)
```
ayeshurun: consider merging

```py
query_value = item.extract_friendly_name_path_or_default(query) # or any other better name :)
```

which will hold 

```py
def get_mutable_prop_path(self, key: str) -> str | None:
        item_type = self.item_type

        if item_type in ITMutablePropMap:
            for prop in ITMutablePropMap[item_type]:
                if key in prop:
                    return prop[key]
        return query
```

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -9,6 +9,20 @@
...
 
+def validate_item_query(query_value: str) -> None:
+    if not (
+        query_value in fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS
+        or query_value == fab_constant.ITEM_QUERY_DEFINITION
+        or query_value.startswith(f"{fab_constant.ITEM_QUERY_DEFINITION}.")
+        or query_value.startswith(f"{fab_constant.ITEM_QUERY_PROPERTIES}.")
+    ):
+        raise FabricCLIError(
+            f"Invalid query '{query_value}'. Allowed queries for items are: {', '.join(fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS)}, "
```
ayeshurun: we should use error messages method instead plain strings here

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -9,6 +9,20 @@
...
+            f"'{fab_constant.ITEM_QUERY_DEFINITION}', '{fab_constant.ITEM_QUERY_DEFINITION}.*', or '{fab_constant.ITEM_QUERY_PROPERTIES}.*'",
+            fab_constant.ERROR_INVALID_INPUT,
```
ayeshurun: do you think it will be clearer for users if we return `ERROR_INVALID_QUERY` or `ERROR_INVALID_QUERY_PARAM`?

src/fabric_cli/core/fab_constant.py
```diff
@@ -316,3 +316,16 @@
...
 }
+
+# Item set constants
+ITEM_QUERY_DEFINITION = "definition"
+ITEM_QUERY_PROPERTIES = "properties"
+ITEM_QUERY_DISPLAY_NAME = "displayName"
+ITEM_QUERY_DESCRIPTION = "description"
+
+# Allowed metadata keys for item set operations
+ITEM_SET_ALLOWED_METADATA_KEYS = [
```
ayeshurun, line 328: what is the reason of maintaing this? 

what about the other types like workspace / capacity / sparkpool. do we need whitelist there as well?

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -16,49 +17,72 @@
...
+        args.id = item.id
+        args.item_uri = format_mapping.get(item.item_type, "items")
 
-        query_value = item.get_property_value(query)
+        if query_value == fab_constant.ITEM_QUERY_DEFINITION or query_value.startswith(
+            f"{fab_constant.ITEM_QUERY_DEFINITION}."
+        ):
+            if not item.check_command_support(Command.FS_EXPORT):
+                raise utils_set.FabricCLIError(
+                    f"Item type '{item.item_type}' does not support definition updates",
```
ayeshurun: same comment here, please define error message method

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -16,49 +17,72 @@
...
+                    utils_set.fab_constant.ERROR_UNSUPPORTED_COMMAND,
```
ayeshurun: why the prefix of utils_set in `utils_set.fab_constant.ERROR_UNSUPPORTED_COMMAND`? can't we use `fab_constant.ERROR_UNSUPPORTED_COMMAND`?

src/fabric_cli/parsers/fab_fs_parser.py
```diff
@@ -419,7 +419,8 @@ def register_set_parser(subparsers: _SubParsersAction) -> None:
         "$ set ws2r.workspace -q sparkSettings.pool.defaultPool -i <inline_json_w_id_name_type> -f",
     ]
     set_learnmore = [
-        "Tip: Run `set <path> -q check -i empty` to see updatable  properties (e.g. set ws1.Workspace -q check -i empty)"
+        "Tip: Run `set <path> -q check -i empty` to see updatable properties (e.g. set ws1.Workspace -q check -i empty)",
```
ayeshurun: do we really want to provide this hacky way ? :)

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -75,6 +94,16 @@ def extract_json_schema(schema: dict, definition: bool = True) -> tuple:
     return definition_properties, name_description_properties
 
 
+def extract_updated_properties(updated_data: dict, query_path: str) -> dict:
```
ayeshurun, line 131: while code wise it is straight forward, can you explain the rationale?

tests/test_commands/test_set.py
```diff
@@ -93,27 +100,23 @@ def test_set_item_report_definition_semantic_model_id_success(
         mock_questionary_print.reset_mock()
         mock_print_done.reset_mock()
         upsert_item_to_cache.reset_mock()
-        property_alias = "semanticModelId"
+
+        property_path = (
+            "definition.parts[0].payload.datasetReference.byConnection.connectionString"
+        )
+
+        new_connection_string = f"Data Source=pbiazure://api.powerbi.com;Initial Catalog={workspace.display_name}/{new_semantic_model.display_name};semanticmodelid={new_semantic_model_id}"
```
ayeshurun, line 114: let's use dummy value here

.changes/unreleased/added-20251113-131349.yaml
```diff
@@ -0,0 +1,3 @@
+kind: added
+body: set item definitions and metadata
```
ayeshurun: consider more meaningful message like `enhance support of set command for items`

tests/test_utils/test_fab_cmd_set_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    resource_def = {"definition": {"parts": [{"x": "old_value"}]}}
+
+    json_string_input = '{"transparency":{"Value":"70D"}}'
+
+    json_payload, updated_def = update_fabric_element(
+        resource_def=resource_def,
+        query="definition.parts[0].x",
+        input=json_string_input,
+        decode_encode=False,
+        raw_string=True,
```
ayeshurun: maybe i asked it already, but can you explain the reason for this raw_string new param? why we need it?

src/fabric_cli/client/fab_api_client.py
may-hartov: From PR description:
Minor bug fix – Call /result API on LRO only when the operation request returns Success and Location header. Otherwise, there is no payload in the result to fetch.

During testing, I observed log entries showing requests to operations/id/result returning 400 with the status operationHasNoResult. This happens because we currently call /result whenever the operation status is Success. The correct approach is to call /result only if the operation response contains a Location header.

src/fabric_cli/core/fab_constant.py
may-hartov: We will probably use something similar per wokrspace / capacity / spark pool. For example - we support setting sku.name for capacities.

I can either hard-code it directly in the setItem method or define it in the constants file. I don’t see any strong advantages or disadvantages to placing it in one location versus the other.

src/fabric_cli/parsers/fab_fs_parser.py
may-hartov: Agreed, I think it should be removed.

tests/test_commands/test_set.py
may-hartov: Dummy value will fail. API will return bad request. E2E tests. This means that if report definition will change this test will start to break on live mode. However I do want to test the "Rebind" flow.

tests/test_utils/test_fab_cmd_set_utils.py
may-hartov: During testing, I found that some properties require a string representation of a JSON object rather than an actual JSON structure. For example, the visualContainers property in report.json expects a value like:
"{\"objects\":{\"background\":[{\"properties\":{\"color\":...}}]}}"

However, the set command uses json.loads to parse the input, which converts the string into a JSON object. This causes the API to fail because it expects a string, not a parsed JSON. To address this, I added a flag for such properties.

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: Previously, the patch Item request included the entire item payload, even unchanged fields such as id, workspaceId, and description, while the API only expects the modified property.

This method extracts the specific target property and its value from the payload. with this method, when the query is "displayName" the request to update item API will include only the displayname in the payload. Note that for "properties" object the request will include the entire properties object, not partial.

src/fabric_cli/client/fab_api_item.py
may-hartov: it's not private. It is also called outside of fab_api_item in fab_fs_set_item

.changes/unreleased/added-20251113-131349.yaml
```diff
@@ -0,0 +1,3 @@
+kind: added
+body: enhance support of set command for items
```
ohadedry: maybe a more detailed description

tests/test_utils/test_fab_cmd_set_utils.py
may-hartov: discussed offline, decided to remove this flag and document that stringified json is currently not supported. We will consider this in future release

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: added tests

src/fabric_cli/core/hiearchy/fab_item.py
```diff
@@ -51,33 +51,14 @@ def job_type(self) -> FabricJobType:
...
-        for prop in properties:
-            if key in prop:
-                return prop[key]
-
-        raise FabricCLIError(
-            ErrorMessages.Hierarchy.key_not_found_in_properties(key),
-            fab_constant.ERROR_INVALID_PROPERTY,
-        )
+            for prop in ITMutablePropMap[item_type]:
+                if key in prop:
```
ayeshurun, line 59: in or equal?

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -6,9 +6,32 @@
...
+            or start with "definition." or "properties.".
+
+    Raises:
+        FabricCLIError: If the query is not in an allowed format.
+    """
+    if not (
+        query_value in fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS
+        or query_value == fab_constant.ITEM_QUERY_DEFINITION
+        or query_value.startswith(f"{fab_constant.ITEM_QUERY_DEFINITION}.")
+        or query_value.startswith(f"{fab_constant.ITEM_QUERY_PROPERTIES}.")
```
ayeshurun: what about a scenario of querying the entire `properties` object? 

like

fab set <item> -q properties -i {prop1: value, prop2: value2}

ayeshurun: btw, why don't we use the validate_expression method below? seems relevant as we just need the allowed keys

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -17,48 +20,81 @@ def exec(item: Item, args: Namespace) -> None:
...
     utils_set.print_set_warning()
     if force or utils_ui.prompt_confirm():
+        args.output = None
+        args.deep_traversal = True
+        args.ws_id = item.workspace.id
+        args.id = item.id
+        args.item_uri = format_mapping.get(item.item_type, "items")
 
-        query_value = item.get_property_value(query)
+        if query_value == fab_constant.ITEM_QUERY_DEFINITION or query_value.startswith(
```
ayeshurun: do you think we can merge this code inside the validation method?

ayeshurun:
```suggestion
        if query_value.startswith(fab_constant.ITEM_QUERY_DEFINITION) 
```

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -17,48 +20,81 @@ def exec(item: Item, args: Namespace) -> None:
...
+    decode_encode: bool,
+) -> tuple[str, dict]:
+    try:
+        return utils_set.update_fabric_element(
+            resource_def,
+            query_value,
+            input_value,
+            decode_encode=decode_encode,
+        )
+    except (ValueError, KeyError, IndexError):
```
ayeshurun, line 87: any reason for these specific errors? simply can be 

```py
except Exception:
```

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -17,48 +20,81 @@ def exec(item: Item, args: Namespace) -> None:
...
-            ):
-                item_api.update_item_definition(args, update_item_definition_payload)
-        except Exception:
-            utils_ui.print_grey(
-                "Item supports only updating displayName or description, not definition",
+            definition_base64_to_update, _ = utils_set.extract_json_schema(updated_def)
+            update_item_definition_payload = json.dumps(definition_base64_to_update)
+
+            utils_ui.print_grey(f"Setting new property for '{item.name}'...")
+            item_api.update_item_definition(args, update_item_definition_payload)
```
ayeshurun, line 48: i see the current code wraps this with try-catch block. why you remove it? (i;m not familiar with any item that supports get definition and not update, i think they always come together, but i just want to make sure i'm not missing anything)

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -17,48 +20,81 @@ def exec(item: Item, args: Namespace) -> None:
...
-            item_def, query_value, args.input, decode_encode=True
-        )
+            args.format = definition_format_mapping.get(item.item_type, "")
+            def_response = item_api.get_item_definition(args)
+            definition = json.loads(def_response.text)
 
-        definition_base64_to_update, name_description_properties = (
-            utils_set.extract_json_schema(updated_def)
-        )
+            json_payload, updated_def = _update_element(
```
ayeshurun: not sure why _update_element returns json_payload. both here and below `json_payload` is not used

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -17,48 +20,81 @@ def exec(item: Item, args: Namespace) -> None:
...
-
-        try:
-            if query_value.startswith("definition") and item.check_command_support(
-                Command.FS_EXPORT
-            ):
-                item_api.update_item_definition(args, update_item_definition_payload)
-        except Exception:
-            utils_ui.print_grey(
-                "Item supports only updating displayName or description, not definition",
+            definition_base64_to_update, _ = utils_set.extract_json_schema(updated_def)
```
ayeshurun, line 44: writing to take in a separate PR - improve this method

src/fabric_cli/utils/fab_cmd_set_utils.py
ayeshurun: P2 for this feature - consider supporting -q as multiple fields

like

`fab set <item> -q {displayName, propperties: {prop1}} -i {displayName: value, properties: {prop1: value1} }`

tests/test_commands/test_set.py
```diff
@@ -93,27 +100,23 @@ def test_set_item_report_definition_semantic_model_id_success(
         mock_questionary_print.reset_mock()
         mock_print_done.reset_mock()
         upsert_item_to_cache.reset_mock()
-        property_alias = "semanticModelId"
```
ayeshurun, line 96: do we want to have a test for the old rebind approach?

tests/test_commands/test_set.py
```diff
@@ -71,14 +78,14 @@ def test_set_item_metadata_success(
             cassette_name,
         )
 
-    @pytest.mark.skip()
     def test_set_item_report_definition_semantic_model_id_success(
```
ayeshurun, line 87: consider adding additional test for properties update (currently only VariableLibrary)

tests/test_utils/test_fab_cmd_set_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+import pytest
+
+from fabric_cli.utils.fab_cmd_set_utils import (
+    extract_updated_properties,
+    update_fabric_element,
+)
+
+
+def test_update_fabric_element_with_json_input():
```
ayeshurun: nit: `sucess` suffix

tests/test_utils/test_fab_cmd_set_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    assert updated_def["definition"]["parts"][0]["x"]["transparency"]["Value"] == "70D"
+
+    parsed_payload = json.loads(json_payload)
+    assert isinstance(parsed_payload["definition"]["parts"][0]["x"], dict)
+    assert (
+        parsed_payload["definition"]["parts"][0]["x"]["transparency"]["Value"] == "70D"
+    )
+
+
+def test_extract_updated_properties_preserves_sibling_properties_success():
```
ayeshurun, line 38: worth adding a test for

```
def test_extract_updated_properties_exclude_redundant_properties_success():
    updated_metadata = {"k0: "value0", "k": {"k1": "v1", "k2": {"k3": "value"}}}
    query_path = "k.k2.k3"

    update_payload_dict = extract_updated_properties(updated_metadata, query_path)

    assert "k" in update_payload_dict
    assert "k0" not in update_payload_dict
    assert update_payload_dict["k"]["k1"] == "v1"
    assert update_payload_dict["k"]["k2"]["k3"] == "value"
```

docs/commands/fs/set.md
```diff
@@ -9,18 +9,82 @@ Set a workspace or item property.
...
+- `properties` - Custom properties (`.VariableLibrary` only)
+
+### 2. Item Definition
+
+Any explicit path (specified via the `-q` / `--query` command argument) to properties within the item's `definition` structure according to [Microsoft Fabric item definitions](https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions).
+
+### Limitations
+
+- Only one property path can be specified per `--query` argument
+- Paths must map directly to JSON paths **without** filters or wildcards
```
ayeshurun, line 47: are we validating this? i guess we should for early failure

docs/commands/fs/set.md
```diff
@@ -9,18 +9,82 @@ Set a workspace or item property.
...
+- Properties can be added one level deep from existing paths returned by [`get`](get.md), provided they are valid according to the item's schema. Nested properties must be created incrementally (e.g., to set `a.b.c`, first set `a.b`, then set `a.b.c`)
```
ayeshurun: why is that?

docs/commands/fs/set.md
```diff
@@ -9,18 +9,82 @@ Set a workspace or item property.
...
+- Properties that expect a JSON string value are not supported - JSON input is always parsed as an object
+
+**Supported - Single property path:**
+
+```bash
```
ayeshurun: consider using some admonitions, as described here:

https://squidfunk.github.io/mkdocs-material/reference/admonitions/

docs/commands/fs/set.md
ayeshurun, line 1: have we validated a draft of the documentation with these changes?

docs/commands/fs/set.md
```diff
@@ -9,18 +9,82 @@ Set a workspace or item property.
...
+### Common Item-Specific Definition Property Paths
+
+These are friendly names that map to specific paths in the item's definition structure. When using the `set` command, you can use these names directly (as the `-query` / `--q` argument value) as they map to the correct definition paths:
+
+- **Notebook**: `lakehouse`, `environment`, `warehouse`
+- **Report**: `semanticModelId` (applies only to [Report definition.pbir version 1](https://learn.microsoft.com/en-us/power-bi/developer/projects/projects-report?tabs=v1%2Cdesktop#definitionpbir). For other versions, check the correct property path in the Report definition documentation)
+- **SparkJobDefinition**: `payload`
+
+!!! note "Note on friendly names"
+    These friendly names may be deprecated in a future release. For forward compatibility, consider using explicit JSON paths within the item's `definition` structure according to [Microsoft Fabric item definitions](https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions).
```
ayeshurun: consider:


These friendly names may be deprecated in a future release. We strongly recommend using explicit JSON paths within the item's `definition` structure according to [Microsoft Fabric item definitions](https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions) rahter than friendly names.

docs/commands/fs/set.md
```diff
@@ -9,18 +9,82 @@ Set a workspace or item property.
...
+    ```powershell
+    fab set item.Resource -q query -i '{\"key\":\"value\"}'
+    ```
+
+=== "Bash/Zsh"
+    ```bash
+    fab set item.Resource -q query -i '{"key":"value"}'
+    ```
+
+## Setting Item Properties
```
ayeshurun: is it the appropriate place for this section? perhaps it should be under Update Item in item_exmaples.md. WDYT?

docs/examples/item_examples.md
```diff
@@ -193,30 +195,49 @@ fab rm ws1.Workspace/lh1.Lakehouse -f
...
+fab set ws1.Workspace/rep1.Report -q definition.parts[0].payload.datasetReference.byConnection.ConnectionString -i "ConnectionStringPrefix....semanticmodelid=00000000-0000-0000-0000-000000000000"
+```
+
+#### Update Notebook Cell Code
+
+Update the code in a specific notebook cell.
+
 ```
-fab set ws1.Workspace/rep1.Report -q semanticModelId -i "00000000-0000-0000-0000-000000000000
+fab set nb1.Notebook -q definition.parts[0].payload.cells[0].source[0] -i "someCode"
```
ayeshurun: not sure about sharing an example of adjusting the code cell (although supported), perhaps we can use something simpler, like 

```py
fab set nb1.Notebook -q definition.parts[0].payload.metadata.dependencies.lakehouse.default_lakehouse -i 00000000-0000-0000-0000-000000000001
```

docs/commands/fs/set.md
may-hartov: I expect set.md to describe the command functionality, structure and limitation while item_exmaples.md is just a usage example. But I don't mind moving it

docs/commands/fs/set.md
may-hartov: ### Single Property Limitation
We’ve already discussed this. For **P0**, we agreed to keep it simple by supporting a single path and revisit multiple arguments later. Note that in the current implementation (before my changes), the `set item` command only accepts one argument value. For P0, if a user wants to update multiple properties in one command, they can do so by providing the query for the first shared ancestor of those properties.

---

### Filters and Wildcards
Regarding **filtering and wildcards**, we also agreed to explore this in the future, starting with usage patterns—what filters make sense for the `set` command? As for wildcards, JMESPath requires users to know the nesting level of the property. For example, `definitions.\*.environment` won’t return the environment path of a notebook; the user would need something like `definitions.\*.\*.\*...\.environment`. This raises questions about whether such usage is practical.

---

### Creation of Nested Properties
Creating nested properties is not straightforward because the query is currently validated against the existing item definition structure. While implementing this might be relatively simple, it requires changes to validation logic. We agreed to examine this in a separate PR.

tests/test_commands/test_set.py
may-hartov: That will force the test to create report from the old definition schema, which I'm not sure how can be achieved

docs/commands/fs/set.md
may-hartov: discussed offline, will rephrase

docs/examples/item_examples.md
may-hartov: notebook lakehouse example already exists. Refactored it to use explicit path rather "lakehouse" name. will remove the set cell example

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
may-hartov: This is not safe. Consider the case someone is typing "definition**s**." instead of "definition" - it will pass the validation you suggested, and call get definitions API. I want to make sure the prefix is either definition or "definition."

tests/test_utils/test_fab_cmd_set_utils.py
may-hartov: discussed offline - this is not the expected functionality - this method is used to get the payload for update item API requst. In case of setting the "properties" field, we expect the entire object in the payload, and not a part of it. we have a test the asserts the method **will** include the siblings

src/fabric_cli/core/hiearchy/fab_item.py
may-hartov: equal is wrong. prop is a key-value pair. for example         {
            "environment": "definition.parts[0].payload.metadata.dependencies.environment"
        }. the "in" operator checks if the key is equal to the prop key.

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: Querying the entire properties will work as "properties" is in ITEM_SET_ALLOWED_METADATA_KEYS. In the validation method we check if query_value in fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS

may-hartov: will refactor to use validate_expression

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
may-hartov: If update item definition API fails, it doesn't mean that the item doesn't support update item definition API. Instead - we will raise the API error in the command and user can see the API error response

tests/test_commands/test_set.py
ayeshurun: don't we have already test for it? with the old schema?

docs/commands/fs/set.md
may-hartov: Fixed documentation and validate draft with local mkdocs

docs/commands/fs/set.md
may-hartov: added jmespath has no filters or wildcards validation

src/fabric_cli/commands/fs/set/fab_fs_set_item.py
may-hartov: these are the expectations utils_jmespath.replace may raise if the query is invalid. While testing the scenario it was not easy to correlate the error thrown from utils_jmespath.replace to the query parameter, so for these exceptions I want to make it clear something with the query is invalid. For other expectations - let it throw.

tests/test_commands/test_set.py
may-hartov: added test_set_item_variable_library_properties_success

docs/commands/fs/set.md
may-hartov: discussed offline, will remove narg+ and validate error handling

tests/test_commands/test_set.py
may-hartov: we had a test marked with "skip" and no record file

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -10,27 +10,43 @@
...
+        FabricCLIError: If the query is not in an allowed format, contains
+            filters/wildcards, or if definition update is not supported for the item type.
     """
-    if not (
-        query_value in fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS
-        or query_value == fab_constant.ITEM_QUERY_DEFINITION
-        or query_value.startswith(f"{fab_constant.ITEM_QUERY_DEFINITION}.")
-        or query_value.startswith(f"{fab_constant.ITEM_QUERY_PROPERTIES}.")
-    ):
+    allowed_keys = fab_constant.ITEM_SET_ALLOWED_METADATA_KEYS + [
```
ayeshurun, line 26: what about ITEM_QUERY_PROPERTIES?

docs/commands/fs/set.md
```diff
@@ -25,66 +25,55 @@ fab set <path> -q <jmespath_query> -i <input_value> [-f]
...
+- If the property path doesn't exist on the item definition, it will be added, provided it's valid according to the item's schema. The `set` command supports creation of 1 level at a time (e.g., to set `a.b.c`, first set `a.b`, then set `a.b.c`)
+- Properties that expect a JSON string value are not supported - JSON input is always parsed as an object
 
-### Best Practice
+## Set Item Support
 
-**surround JSON input with single quotes (`'`). In PowerShell escape inner double quotes with backslashes (`\"`)**
+### Input
+
+When providing JSON input in command-line mode, different shells process quotes and escape characters before passing them to the CLI.
```
ayeshurun: this is not specifically to item / set. maybe move to a common

docs/commands/fs/set.md
may-hartov: discussed offline - created a table mapping supported query values to fabric resource type

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: This is a part of ITEM_SET_ALLOWED_METADATA_KEYS - it's in the item's metadata

docs/commands/fs/set.md
may-hartov: mvoed to essentials/parameters
image

PR #69, opened 2025-11-19
fix(LRO): remove redundant call to /operations/{id}/result when no location header is present

may-hartov:
**Summary**
* Updated the `_poll_operation` function to check for the presence of a `Location` header before fetching operation results, ensuring correct handling of operation responses. If the `Location` header is not present, the response status code is set to 200 and the original response is returned.

**Context**
When the CLI receives a success response from the LRO API, it automatically calls /operations/{id}/result to retrieve the result payload. However, this payload is only available when the operation returns success and includes a Location header in the response. If the header is missing, the /result call is unnecessary and will return the error operationHasNoResult.

### DIFF
--- a/.changes/unreleased/fixed-20251119-095159.yaml
+++ b/.changes/unreleased/fixed-20251119-095159.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: LRO polling - call operations/id/result API only if the location header is present
+time: 2025-11-19T09:51:59.752824783Z


src/fabric_cli/client/fab_api_client.py
```diff
@@ -373,9 +373,14 @@ def _poll_operation(
...
-                        return _fetch_operation_result(
-                            args, uri, response, original_response
-                        )
+                        location_header = response.headers.get("Location", "")
+                        if location_header:
+                            return _fetch_operation_result(
+                                args, uri, response, original_response
+                            )
+
+                        original_response.status_code = 200
```
aliabufoul, line 382: is it possible that we hit a case where the location is not defined, but result status code is not 200, like 400 or so?
If yes, returning 200 by default as a fixed value might not be correct always

ayeshurun: we get location header only if the status code was 202. if the api failed, we don't get the header. but i think that if we got to this code path, the first call was successful (202) and the polling call was also successful (200) with a header that points to the result.

may-hartov: The status response was 200 with status "Succeeded" or "Completed" the original request returned 202. inside _fetch_operation_result we return 200 in case the operation has "OperationHasNoResult" status code.

aviatco:
@may-hartov  can you please add unit test verify this case? that we skip the second call in case no headers

PR #91, opened 2025-12-11
feat: inform user about cli updates

### LINKED ISSUE #71: [FEATURE] Help me keep my fabric-cli updated
### Use Case / Problem

I forgot to update to the latest version of the CLI

### Proposed Solution

Similar to pip install, I could get a friendly reminded that there is a new version of the CLI and the command to upgrade.

Image

To avoid annoying the user, this should only appear once per CLI session.

### Alternatives Considered

_No response_

### Impact Assessment

- [x] This would help me personally
- [x] This would help my team/organization
- [ ] This would help the broader fabric-cli community
- [ ] This aligns with Microsoft Fabric roadmap items

### Implementation Attestation

- [x] I understand this feature should maintain backward compatibility with existing commands
- [x] I confirm this feature request does not introduce performance regressions for existing workflows
- [x] I acknowledge that new features must follow fabric-cli's established patterns and conventions

### Implementation Notes

_No response_

Guust-Franssens:
# 📥 Pull Request
- Resolves #71 

Implements automatic update notifications that inform users when a new version of the Fabric CLI is available on PyPI. The check runs on login (`fab auth login`) and displays a friendly notification if an update is available.

#### New Files:
- `src/fabric_cli/utils/fab_version_check.py` - Core version checking module with PyPI integration
- `tests/test_utils/test_fab_version_check.py` - 100% coverage on fab_version_check.py

#### Modified Files:
- `src/fabric_cli/core/fab_constant.py` - Added constants for update checking configuration
- `src/fabric_cli/commands/auth/fab_auth.py` - Integrated version check on successful login
- `docs/essentials/settings.md` - Documented the new `check_updates` setting

### Implementation Details

- **Once per session**: version check is ran whenever a user logs in using `fab auth login`
- **User Control**: Can be disabled via `fab config set check_updates false`
- **Silent Failures**: Doesn't interrupt user experience if PyPI is unreachable
- **Conservative Version Comparison**: Uses tuple-based semantic versioning without external dependencies
- **Debug Logging**: Integrated with existing `fab_logger` for troubleshooting

### Configuration
New config key `check_cli_version_updates` (default: `true`) stored in `~/.config/fab/config.json`. This setting enables/disables update notifications.

### User Experience
When a user logs in and a new version is available:
```bash
$ fab auth login -u "$CLIENT_ID" -p "$CLIENT_SECRET" --tenant "$TENANT_ID"

[notice] A new release of fab is available: 1.1.0 → 1.2.0
[notice] To update, run: pip install --upgrade ms-fabric-cli
```

### DIFF
--- a/src/fabric_cli/utils/fab_version_check.py
+++ b/src/fabric_cli/utils/fab_version_check.py
@@ -31,8 +31,9 @@ def _fetch_latest_version_from_pypi() -> Optional[str]:
         )
         response.raise_for_status()
         return response.json()["info"]["version"]
-    except (requests.RequestException, KeyError, ValueError, TypeError):
+    except (requests.RequestException, KeyError, ValueError, TypeError) as e:
         # Silently fail - don't interrupt user experience for version checks
+        fab_logger.log_debug(f"Failed to fetch version from PyPI: {e}")
         return None
 
 
--- a/tests/test_utils/test_fab_version_check.py
+++ b/tests/test_utils/test_fab_version_check.py
@@ -68,19 +68,25 @@ def test_cli_version_fetch_pypi_success(mock_get):
     )
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check.requests.get")
-def test_cli_version_fetch_pypi_network_error_failure(mock_get):
+def test_cli_version_fetch_pypi_network_error_failure(mock_get, mock_log_debug, mock_fab_set_state_config):
     """Should return None when PyPI request fails (network error)."""
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_get.side_effect = requests.ConnectionError("Network error")
 
     result = fab_version_check._fetch_latest_version_from_pypi()
 
     assert result is None
+    mock_log_debug.assert_called_once()
+    assert "Failed to fetch version from PyPI" in str(mock_log_debug.call_args[0][0])
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check.requests.get")
-def test_cli_version_fetch_pypi_http_error_failure(mock_get):
+def test_cli_version_fetch_pypi_http_error_failure(mock_get, mock_log_debug, mock_fab_set_state_config):
     """Should return None when PyPI returns non-200 status code."""
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_response = MagicMock()
     mock_response.status_code = 500
     mock_response.raise_for_status.side_effect = requests.HTTPError("500 Server Error")
@@ -89,11 +95,15 @@ def test_cli_version_fetch_pypi_http_error_failure(mock_get):
     result = fab_version_check._fetch_latest_version_from_pypi()
 
     assert result is None
+    mock_log_debug.assert_called_once()
+    assert "Failed to fetch version from PyPI" in str(mock_log_debug.call_args[0][0])
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check.requests.get")
-def test_cli_version_fetch_pypi_invalid_json_failure(mock_get):
+def test_cli_version_fetch_pypi_invalid_json_failure(mock_get, mock_log_debug, mock_fab_set_state_config):
     """Should return None when PyPI returns invalid JSON."""
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_response = MagicMock()
     mock_response.status_code = 200
     mock_response.raise_for_status.return_value = None
@@ -103,11 +113,15 @@ def test_cli_version_fetch_pypi_invalid_json_failure(mock_get):
     result = fab_version_check._fetch_latest_version_from_pypi()
 
     assert result is None
+    mock_log_debug.assert_called_once()
+    assert "Failed to fetch version from PyPI" in str(mock_log_debug.call_args[0][0])
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check.requests.get")
-def test_cli_version_fetch_pypi_missing_keys_failure(mock_get):
+def test_cli_version_fetch_pypi_missing_keys_failure(mock_get, mock_log_debug, mock_fab_set_state_config):
     """Should return None when PyPI response is missing expected keys."""
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_response = MagicMock()
     mock_response.status_code = 200
     mock_response.raise_for_status.return_value = None
@@ -117,6 +131,8 @@ def test_cli_version_fetch_pypi_missing_keys_failure(mock_get):
     result = fab_version_check._fetch_latest_version_from_pypi()
 
     assert result is None
+    mock_log_debug.assert_called_once()
+    assert "Failed to fetch version from PyPI" in str(mock_log_debug.call_args[0][0])
 
 
 def test_cli_version_compare_newer_major_success():
@@ -192,40 +208,49 @@ def test_cli_version_check_new_version_available_success(
     assert "pip install --upgrade" in call_msg
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check._fetch_latest_version_from_pypi")
 def test_cli_version_check_same_version_success(
-    mock_fetch, mock_fab_set_state_config, mock_questionary_print
+    mock_fetch, mock_log_debug, mock_fab_set_state_config, mock_questionary_print
 ):
     """Should not display notification when on latest version."""
     mock_fab_set_state_config(fab_constant.FAB_CHECK_UPDATES, "true")
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_fetch.return_value = __version__
 
     fab_version_check.check_and_notify_update()
 
     mock_questionary_print.assert_not_called()
+    mock_log_debug.assert_called_with(f"Already on latest version: {__version__}")
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check._fetch_latest_version_from_pypi")
 def test_cli_version_check_fetch_failure(
-    mock_fetch, mock_fab_set_state_config, mock_questionary_print
+    mock_fetch, mock_log_debug, mock_fab_set_state_config, mock_questionary_print
 ):
     """Should not display notification when PyPI fetch fails."""
     mock_fab_set_state_config(fab_constant.FAB_CHECK_UPDATES, "true")
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_fetch.return_value = None  # Simulate fetch failure
 
     fab_version_check.check_and_notify_update()
 
     mock_questionary_print.assert_not_called()
+    mock_log_debug.assert_called_with("Could not fetch latest version from PyPI")
 
 
+@patch("fabric_cli.core.fab_logger.log_debug")
 @patch("fabric_cli.utils.fab_version_check._fetch_latest_version_from_pypi")
 def test_cli_version_check_older_version_success(
-    mock_fetch, mock_fab_set_state_config, mock_questionary_print
+    mock_fetch, mock_log_debug, mock_fab_set_state_config, mock_questionary_print
 ):
     """Should not display notification when PyPI version is older."""
     mock_fab_set_state_config(fab_constant.FAB_CHECK_UPDATES, "true")
+    mock_fab_set_state_config(fab_constant.FAB_DEBUG_ENABLED, "true")
     mock_fetch.return_value = _decrement_version()
 
     fab_version_check.check_and_notify_update()
 
     mock_questionary_print.assert_not_called()
+    mock_log_debug.assert_called_with(f"Already on latest version: {__version__}")


src/fabric_cli/core/fab_constant.py
```diff
@@ -111,6 +119,9 @@
     FAB_OUTPUT_FORMAT: ["text", "json"],
     FAB_FOLDER_LISTING_ENABLED: ["false", "true"],
     FAB_WS_PRIVATE_LINKS_ENABLED: ["false", "true"],
+    FAB_CHECK_UPDATES: ["false", "true"],
+    FAB_PYPI_CACHE_TIMESTAMP: [],
+    FAB_PYPI_LATEST_VERSION: [],
```
ayeshurun: @Guust-Franssens could you please explain why it should be in config.json and not in auth.json? the config are keys that we loop over and return in `fab config ls`. in addition, most of the values are cleaned up when the user logging out. is that the behavior we want?

src/fabric_cli/core/fab_constant.py
```diff
@@ -91,6 +91,14 @@
 FAB_OUTPUT_FORMAT = "output_format"
 FAB_FOLDER_LISTING_ENABLED = "folder_listing_enabled"
 FAB_WS_PRIVATE_LINKS_ENABLED = "workspace_private_links_enabled"
+FAB_CHECK_UPDATES = "check_updates"
+FAB_PYPI_CACHE_TIMESTAMP = "pypi_cache_timestamp"
+FAB_PYPI_LATEST_VERSION = "pypi_latest_version"
+
+# Version check settings
+VERSION_CHECK_INTERVAL_HOURS = 24 * 7  # Check once a week
```
aviatco: I think a 7‑day interval is too long. For example, if we check the value one day before the release, the message might only appear six days later.
Also, why do we need to keep these variables: FAB_PYPI_CACHE_TIMESTAMP and FAB_PYPI_LATEST_VERSION?
In my opinion, we don’t need to store these values. Each time a user logs in, we can simply perform the validation, it’s not a heavy operation, and users don’t log in frequently anyway.
This approach will also simplify the logic and the code.

.changes/unreleased/added-20251211-183425.yaml
```diff
@@ -0,0 +1,3 @@
+kind: added
+body: inform users whenever a new update is available
```
aviatco: consider rephrase: Display a notification to users on login when a new version is available

src/fabric_cli/core/fab_constant.py
Guust-Franssens: I agree let's keep it simple. During my brief research into how pip does it (original feature request mentions pip) I saw that they were [caching it](https://github.com/pypa/pip/blob/9de73b0c547943568d6ab97b7c62a4d2d94c2501/src/pip/_internal/self_outdated_check.py#L84-L138) and without much thought of whether it is needed, I copied their approach.

.changes/unreleased/added-20251211-183425.yaml
Guust-Franssens: changed it to your suggestion

src/fabric_cli/core/fab_constant.py
Guust-Franssens: Hey @ayeshurun. I removed the caching mechanism (see @aviatco 's remark). This makes the code simpler and removes this issue of having it show up in fab config ls.

src/fabric_cli/utils/fab_version_check.py
```diff
@@ -0,0 +1,90 @@
...
+        fab_logger.log_debug("Version check disabled by user configuration")
+        return
+
+    # Check PyPI for latest version
+    fab_logger.log_debug("Checking PyPI for latest version")
+    latest_version = _fetch_latest_version_from_pypi()
+
+    # Display notification if update available
+    if latest_version and _is_pypi_version_newer(latest_version):
+        fab_logger.log_debug(f"New version available: {latest_version}")
```
aviatco: I think this is redundant.  fab_logger.log_debug print to the UI console and we already print the same in the next line

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,230 @@
...
+        return f"{parts[0]}.{parts[1] - 1}.0"
+    # If major > 0, decrement major and set others to 0
+    elif parts[0] > 0:
+        return f"{parts[0] - 1}.0.0"
+    # Edge case: version is 0.0.0 - can't go lower, just return it
+    return "0.0.0"
+
+
+class TestFetchLatestVersionFromPyPI:
+    """Test _fetch_latest_version_from_pypi logic."""
```
aviatco: Please combine the classes to one TestVersionCheck

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @patch("fabric_cli.utils.fab_version_check.requests.get")
+    def test_fetch_success(self, mock_get):
```
aviatco: Please use more meaningfull name such as: test_cli_version_fetch_success / test_cli_version_fetch_failure

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,230 @@
...
+        """Should return None when PyPI request fails (network error)."""
+        mock_get.side_effect = requests.ConnectionError("Network error")
+
+        result = fab_version_check._fetch_latest_version_from_pypi()
+
+        assert result is None
+
+    @patch("fabric_cli.utils.fab_version_check.requests.get")
+    def test_fetch_failure_http_error(self, mock_get):
+        """Should return None when PyPI returns non-200 status code."""
```
aviatco: please align to the naming convention and use failure/success as suffix.

In other tests as well

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,230 @@
...
+        """Should display notification when newer version is available."""
+        newer_version = _increment_version("major")
+        mock_get_config.return_value = "true"
+        mock_fetch.return_value = newer_version
+
+        with patch("fabric_cli.utils.fab_version_check.fab_ui") as mock_ui:
+            fab_version_check.check_and_notify_update()
+
+            # Should display notification
+            mock_ui.print_grey.assert_called()
```
aviatco: you can use mock_questionary_print from tests/conftest.py instead of mock_ui.print_grey

same comment for other tests

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,230 @@
...
+            call[0][0] == fab_constant.FAB_CHECK_UPDATES
+            for call in mock_get_config.call_args_list
+        )
+
+    @patch("fabric_cli.utils.fab_version_check._fetch_latest_version_from_pypi")
+    @patch("fabric_cli.utils.fab_version_check.fab_state_config.get_config")
+    def test_new_version_available(self, mock_get_config, mock_fetch):
+        """Should display notification when newer version is available."""
+        newer_version = _increment_version("major")
+        mock_get_config.return_value = "true"
```
aviatco: You can use  mock_fab_set_state_config(fab_constant.FAB_CHECK_UPDATES, "true") instead

same comment for other tests

src/fabric_cli/utils/fab_version_check.py
Guust-Franssens: removed in 193bc91b6ccb69495c0be9ef51f7302083e92853

tests/test_utils/test_fab_version_check.py
Guust-Franssens: I will leave out the use of test classes entirely. I see that other test files also do not use them unless there is a reason for it (shared fixtures or setup/teardown). In my case I did more of a logical grouping of different parts but I think it makes less sense now especially since the caching part is removed.

tests/test_utils/test_fab_version_check.py
Guust-Franssens: resolved in ccf89bff7f86300d4efca58b4f77b2aea7a1c276

tests/test_utils/test_fab_version_check.py
Guust-Franssens: Resolved in f3eb025b403886ffa4680c287cb522241422d3d8

tests/test_utils/test_fab_version_check.py
Guust-Franssens: done in 8e3e69e896ffcf50a9855f2a6ad18736b0b002e2

tests/test_utils/test_fab_version_check.py
Guust-Franssens: done in 8e3e69e896ffcf50a9855f2a6ad18736b0b002e2

Guust-Franssens:
Thanks for the review @aviatco. Your points are valid and I will keep them in mind in my potential future PRs.

I left some of the remarks open (like proper naming and the class one) as these warrant a second review by you to verify if my changes are meeting your request.

Guust-Franssens:
@aviatco @ayeshurun in f4b4c4e607171997fc6e2ca7e7c944d72ca67d4e I made a change to ignore pypi.org in the VCR playback. This was causing issues in the `tests/test_commands/test_auth.py`.

This passes the tests however now during testing of the fab auth login, actual requests to PyPI are being made. I think it would be cleaner to either:

- Re-record the requests made to include PyPI (preferable by one of you as I don't have the full setup)
- I mock the function making the call to PyPI for all tests to `auth login`. Here I would just make it the mock return value equal to the current `__version__` 

Could you provide some guidance here?

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -199,7 +199,11 @@ def init(args: Namespace) -> Any:
         except KeyboardInterrupt:
             # User cancelled the authentication process
             return False
-        return True
+
+    # Check for CLI updates after successful authentication
```
aviatco: Please avoid comments that explain logic which is already clear from the code itself

src/fabric_cli/utils/fab_version_check.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Check for CLI updates and display notification if a newer version is available.
+
+    This function:
+    - Respects user's check_updates config setting
+    - Checks PyPI on every login for the latest version
+    - Displays notification if an update is available
+    - Fails silently if PyPI is unreachable
+    """
+    # Check if user has disabled update checks
```
aviatco: Please avoid comments that explain logic which is already clear from the code itself. 
Please remove the comments elsewhere too.

src/fabric_cli/utils/fab_version_check.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # Display notification if update available
+    if latest_version and _is_pypi_version_newer(latest_version):
+        msg = (
+            f"\n[notice] A new release of fab is available: {__version__} → {latest_version}\n"
+            "[notice] To update, run: pip install --upgrade ms-fabric-cli\n"
+        )
+        fab_ui.print_grey(msg)
+    elif latest_version:
+        fab_logger.log_debug(f"Already on latest version: {__version__}")
```
aviatco, line 85: why did you used fab_logger.log_debug and not fab_ui.print_grey? using fab_logger.log_debug  means user will see it only if fab_constant.FAB_DEBUG_ENABLED is set by the user to true.

Also, are those scenarios tested?

tests/test_utils/test_fab_version_check.py
```diff
@@ -0,0 +1,229 @@
...
+    mock_fetch, mock_fab_set_state_config, mock_questionary_print
+):
+    """Should display notification when newer version is available."""
+    newer_version = _increment_version("major")
+    mock_fab_set_state_config(fab_constant.FAB_CHECK_UPDATES, "true")
+    mock_fetch.return_value = newer_version
+
+    fab_version_check.check_and_notify_update()
+
+    # Should display notification
```
aviatco: remove unnecessary comment - it is clear from the test's name.
please avoid using it elsewhere

src/fabric_cli/commands/auth/fab_auth.py
Guust-Franssens: resolved in 14742d2d3597f13268c76f00fbbabac130fbae91

src/fabric_cli/utils/fab_version_check.py
Guust-Franssens: resolved in 14742d2d3597f13268c76f00fbbabac130fbae91

tests/test_utils/test_fab_version_check.py
Guust-Franssens: resolved in 14742d2d3597f13268c76f00fbbabac130fbae91

src/fabric_cli/utils/fab_version_check.py
Guust-Franssens: I'm using `fab_logger.log_debug` here intentionally for the "already up to date" scenario. The rationale is:
- Silent success when the user is already on the latest version.
- User-facing notification only when actionable (in case of a newer version available)

100% coverage on `fab_version_check.py`:
```
Name                                        Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------------
src/fabric_cli/utils/fab_version_check.py      32      0      6      0   100%
---------------------------------------------------------------------------------------
TOTAL                                          32      0      6      0   100% 
```

aviatco: agree for the use of fab_logger.log_debug
regarding tests - my Q was that if we verify the correct message is printed only when debug_enabled is true

Guust-Franssens: It was already tested before by this function, however I made it a bit more explicit by adding `mock_questionary_print.assert_not_called()`. 

https://github.com/microsoft/fabric-cli/blob/b56ef8e2ede34150a4cf53827768e401ceb6638b/tests/test_utils/test_fab_version_check.py#L164-L175

src/fabric_cli/core/fab_constant.py
```diff
@@ -91,6 +91,11 @@
 FAB_OUTPUT_FORMAT = "output_format"
 FAB_FOLDER_LISTING_ENABLED = "folder_listing_enabled"
 FAB_WS_PRIVATE_LINKS_ENABLED = "workspace_private_links_enabled"
+FAB_CHECK_UPDATES = "check_updates"
```
aviatco:
```suggestion
FAB_CHECK_UPDATES = "check_cli_version_updates"
```

src/fabric_cli/utils/fab_version_check.py
aviatco: What’s missing is a check that the correct message is logged when debug_enabled=True in test_cli_version_check_same_version_success. You verified mock_questionary_print.assert_not_called(), but you should also assert that log.debug was called with the expected message.

use mock_fab_set_state_config(constant.FAB_DEBUG_ENABLED, True) to set debug_enabled=True then mock fab_logget.debug and verify the message


Same for: "test_cli_version_check_fetch_failure" and "test_cli_version_check_older_version_success"

Guust-Franssens: Apologies, I didn't know that you were referring to testing the logger outputs. I fixed it in b384c46e815dd10e218bf546cbf5dff687254348

aviatco (approved):

PR #100, opened 2025-12-24
fix: create singleton decorator

aviatco:
# 📥 Pull Request

To enable reuse, refactor the singleton logic currently in fab_context into a decorator and apply that decorator within fab_interactive.

### DIFF
--- a/src/fabric_cli/core/fab_decorators.py
+++ b/src/fabric_cli/core/fab_decorators.py
@@ -21,9 +21,6 @@ def singleton(class_):
         if class_ not in instances:
             instances[class_] = class_(*args, **kwargs)
         return instances[class_]
-
-    # Add a reset method to the decorator function for testing purposes
-    getinstance.reset_instance = lambda: instances.pop(class_, None)
     
     return getinstance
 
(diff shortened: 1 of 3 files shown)


src/fabric_cli/core/fab_decorators.py
```diff
@@ -9,11 +9,25 @@
...
+def singleton(class_):
+    """Singleton decorator that ensures only one instance of a class exists."""
+    instances = {}
+
+    def getinstance(*args, **kwargs):
+        if class_ not in instances:
+            instances[class_] = class_(*args, **kwargs)
+        return instances[class_]
+
+    # Add a reset method to the decorator function for testing purposes
```
ayeshurun: if looks like we don't use reset_instance in code, only in tests. if so, let's remove it :) if necessary we will add in the future

aviatco: we use it in the tests: test_config and test_fab_interactiveCLI

aviatco: This was only used in tests, so I removed it. I don’t think the code should contain test‑specific logic.

PR #105, opened 2025-12-28
fix(set): set onelake shortcut with query target bug fix

may-hartov:
The set shortcut command fails when using -q target with the following error:
set: [InvalidInput] The new name matches the existing name. No changes will be made

**Root Cause**
When the name property is not the target of the set shortcut command, the "name" of the shortcut remains the same - and the operation will fail (in fab_fs_set_onelake), even though the supported queries defined in JMESPATH_UPDATE_SHORTCUT include target.

Updating the shortcut target is a valid and useful scenario, for example when redirecting a shortcut to a different lakehouse 

**Fix**
When the command query is target, the CLI now calls the Create Shortcut API with [ShortcutConflictPolicy=OverwriteOnly](https://learn.microsoft.com/en-us/rest/api/fabric/core/onelake-shortcuts/create-shortcut?tabs=HTTP#shortcutconflictpolicy).
Since OneLake shortcuts do not expose a public Update API, updating a shortcut is performed by creating it again with the same name and path while using the OverwriteOnly conflict policy, which replaces the existing shortcut when the name and path are unchanged.
This behavior aligns with the documented REST API semantics. [OneLake Sh...soft Learn | External]
Validation

Added test coverage for the target update scenario.


src/fabric_cli/commands/fs/set/fab_fs_set_onelake.py
```diff
@@ -29,29 +29,20 @@ def onelake_shortcut(onelake: OneLakeItem, args: Namespace) -> None:
...
         new_name = updated_def.get("name", "")
-        if new_name == current_name:
-            raise FabricCLIError(
-                f"The new name matches the existing name. No changes will be made",
-                fab_constant.ERROR_INVALID_INPUT,
-            )
-
-        if "target" in updated_def and "type" in updated_def["target"]:
-            del updated_def["target"]["type"]
-        json_payload = updated_def
```
ayeshurun, line 47: don't we need it?

may-hartov: We don’t actually need this. The API ignores unsupported properties and still successfully applies the target query.

That said, I already have another PR where I extract only the --query path from the updated object and use it as the update payload. I can move this cleanup there, where it’s more relevant.

PR #109, opened 2025-12-31
feat: Switching to interactive when running "fab"

aviatco:
# 📥 Pull Request

### 🎯 Overview
This PR implements automatic REPL (Read-Eval-Print-Loop) functionality by making the Fabric CLI automatically enter interactive mode when users run the fab command without arguments, eliminating the need for manual mode configuration.

#### Auto-Interactive Mode Entry
Add logic to detect when fab is run without arguments and automatically launch interactive mode
Enhanced Error Handling: Refactored main error handling with dedicated functions _handle_keyboard_interrupt() and _handle_unexpected_error() for better separation of concerns
Graceful Startup: Interactive mode now starts automatically with proper exception handling and user feedback
Interactive Mode Improvements
#### Enhanced interactive CLI with better command handling:
Added specific handling for fab command within interactive mode (shows helpful message instead of error)
Improved empty input handling to gracefully continue session
Better error handling for SystemExit and unexpected exceptions in command execution
Enhanced keyboard interrupt (Ctrl+C) and EOF handling with informative messages
More user-friendly error messages for invalid commands
Mode Configuration Deprecation
#### Added deprecation warnings for mode configuration:
Shows warning that mode configuration is deprecated when users try to set interactive/command-line mode
Maintains backward compatibility while guiding users toward the new auto-interactive behavior
Improved messaging for mode transitions
Constants and Configuration
Removed "fab" from interactive help commands list to prevent conflicts with the new auto-interactive functionality

### 🔄 User Experience Improvements
Simplified Workflow: Users can now simply run fab to enter interactive mode
Better Error Messages: More helpful guidance when commands fail or are invalid in interactive mode
Graceful Interruption Handling: Improved Ctrl+C behavior that doesn't exit the session immediately
Backward Compatibility: Existing mode configuration still works with deprecation guidance
### 🎨 Technical Implementation
Auto-Detection Logic: Main entry point detects when no command is provided and launches interactive mode
Error Isolation: Separated error handling functions for cleaner main() function
Enhanced Interactive Loop: Better command parsing and error recovery in interactive sessions
Singleton Pattern: Maintains proper InteractiveCLI singleton behavior
### ✅ Breaking Changes
None - this is a backward-compatible enhancement that adds functionality while maintaining existing behavior.

### DIFF
--- a/.changes/unreleased/added-20251211-183425.yaml
+++ b/.changes/unreleased/added-20251211-183425.yaml
@@ -1,0 +1,3 @@
+kind: added
+body: Display a notification to users on login when a new fab cli version is available
+time: 2025-12-11T18:34:25.601088227+01:00
--- a/.changes/unreleased/fixed-20260102-172853.yaml
+++ b/.changes/unreleased/fixed-20260102-172853.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Fix a typo in connection roles
+time: 2026-01-02T17:28:53.418628486+01:00
--- a/docs/essentials/settings.md
+++ b/docs/essentials/settings.md
@@ -7,6 +7,7 @@ The Fabric CLI provides a comprehensive set of configuration settings that allow
 | Name                           | Description                                                                                 | Type       | Default |
 |--------------------------------|-------------------------------------------------------------------------------------------- |------------|---------|
 | `cache_enabled`                | Toggles caching of CLI HTTP responses                           | `BOOLEAN`  | `true`  |
+| `check_cli_version_updates`                | Enables automatic update notifications on login                                            | `BOOLEAN`  | `true`  |
 | `debug_enabled`                | Toggles additional diagnostic logs for troubleshooting                                   | `BOOLEAN`  | `false` |
 | `context_persistence_enabled`  | Persists CLI navigation context in command line mode across sessions | `BOOLEAN` | `false` |
 | `encryption_fallback_enabled`  | Permits storing tokens in plain text if secure encryption is unavailable                  | `BOOLEAN`  | `false` |
--- a/src/fabric_cli/commands/acls/fab_acls_set.py
+++ b/src/fabric_cli/commands/acls/fab_acls_set.py
@@ -17,7 +17,7 @@ from fabric_cli.utils import fab_ui as utils_ui
 
 # Valid roles for different contexts
 WORKSPACE_ROLES = ["Admin", "Member", "Contributor", "Viewer"]
-CONNECTION_ROLES = ["Owner", "IserWithReshare", "User"]
+CONNECTION_ROLES = ["Owner", "UserWithReshare", "User"]
 GATEWAY_ROLES = ["Admin", "ConnectionCreator", "ConnectionCreatorWithResharing"]
 PRINCIPAL_TYPES = [
     "User",
--- a/src/fabric_cli/commands/auth/fab_auth.py
+++ b/src/fabric_cli/commands/auth/fab_auth.py
@@ -10,7 +10,7 @@ from fabric_cli.core.fab_context import Context
 from fabric_cli.core.fab_exceptions import FabricCLIError
 from fabric_cli.errors import ErrorMessages
 from fabric_cli.utils import fab_mem_store as utils_mem_store
-from fabric_cli.utils import fab_ui
+from fabric_cli.utils import fab_ui, fab_version_check
 
 
 def init(args: Namespace) -> Any:
@@ -199,7 +199,10 @@ def init(args: Namespace) -> Any:
         except KeyboardInterrupt:
             # User cancelled the authentication process
             return False
-        return True
+
+    fab_version_check.check_and_notify_update()
+
+    return True
 
 
 def logout(args: Namespace) -> None:
--- a/src/fabric_cli/core/fab_constant.py
+++ b/src/fabric_cli/core/fab_constant.py
@@ -91,6 +91,11 @@ FAB_OUTPUT_ITEM_SORT_CRITERIA = "output_item_sort_criteria"
 FAB_OUTPUT_FORMAT = "output_format"
 FAB_FOLDER_LISTING_ENABLED = "folder_listing_enabled"
 FAB_WS_PRIVATE_LINKS_ENABLED = "workspace_private_links_enabled"
+FAB_CHECK_UPDATES = "check_cli_version_updates"
+
+# Version check settings
+VERSION_CHECK_PYPI_URL = "https://pypi.org/pypi/ms-fabric-cli/json"
+VERSION_CHECK_TIMEOUT_SECONDS = 3
 
 FAB_CONFIG_KEYS_TO_VALID_VALUES = {
     FAB_CACHE_ENABLED: ["false", "true"],
@@ -111,6 +116,7 @@ FAB_CONFIG_KEYS_TO_VALID_VALUES = {
     FAB_OUTPUT_FORMAT: ["text", "json"],
     FAB_FOLDER_LISTING_ENABLED: ["false", "true"],
     FAB_WS_PRIVATE_LINKS_ENABLED: ["false", "true"],
+    FAB_CHECK_UPDATES: ["false", "true"],
     # Add more keys and their respective allowed values as needed
 }
 
@@ -127,6 +133,7 @@ CONFIG_DEFAULT_VALUES = {
     FAB_OUTPUT_FORMAT: "text",
     FAB_FOLDER_LISTING_ENABLED: "false",
     FAB_WS_PRIVATE_LINKS_ENABLED: "false",
+    FAB_CHECK_UPDATES: "true",
 }
 
 # Command descriptions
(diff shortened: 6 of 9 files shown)


.changes/unreleased/added-20251231-171025.yaml
```diff
@@ -0,0 +1,3 @@
+kind: added
+body: Support REPL- Switch to interactive when runnig fab command
```
ayeshurun: Support `fab` command to start interactive (repl) mode

src/fabric_cli/core/fab_interactive.py
```diff
@@ -61,6 +61,12 @@ def handle_command(self, command):
         elif command in fab_constant.INTERACTIVE_VERSION_COMMANDS:
             utils_ui.print_version()
             return False  # Do not exit
+        elif command.strip() == "fab":
+            utils_ui.print("You are already in interactive mode. Type 'help' for available commands.")
```
ayeshurun: maybe other text like `You don’t need to prefix commands with fab when you’re in interactive mode. Just type --help to see the list of available commands.`

src/fabric_cli/core/fab_interactive.py
```diff
@@ -61,6 +61,12 @@ def handle_command(self, command):
...
+            return False  # Stay in interactive mode
+        elif not command.strip():
+            # Handle empty input gracefully - just continue
```
ayeshurun: let's remove all unnecessary comments

src/fabric_cli/core/fab_interactive.py
```diff
@@ -87,11 +93,19 @@ def handle_command(self, command):
                         utils_ui.print(
                             f"No function associated with the command: {command.strip()}"
                         )
-                except SystemExit:
+                except SystemExit as e:
                     # Catch SystemExit raised by ArgumentParser and prevent exiting
+                    # This handles cases like --help or invalid arguments in interactive mode
+                    if e.code != 0:
+                        utils_ui.print(f"Command failed with exit code: {e.code}")
```
ayeshurun: not sure what value it brings. users should interact with messages and text in interactive mode.

src/fabric_cli/core/fab_interactive.py
```diff
@@ -87,11 +93,19 @@ def handle_command(self, command):
...
+                    return
+                except Exception as e:
+                    # Handle unexpected errors in command execution
+                    utils_ui.print(f"Error executing command: {str(e)}")
                     return
             else:
-                self.parser.error(f"invalid choice: '{command.strip()}'")
+                # Handle invalid commands more gracefully in interactive mode
```
ayeshurun: please remove similar comments

src/fabric_cli/core/fab_interactive.py
```diff
@@ -87,11 +93,19 @@ def handle_command(self, command):
...
+                except SystemExit as e:
                     # Catch SystemExit raised by ArgumentParser and prevent exiting
+                    # This handles cases like --help or invalid arguments in interactive mode
+                    if e.code != 0:
+                        utils_ui.print(f"Command failed with exit code: {e.code}")
+                    return
+                except Exception as e:
+                    # Handle unexpected errors in command execution
+                    utils_ui.print(f"Error executing command: {str(e)}")
                     return
```
ayeshurun, line 100: is't it already covered in fab_exceptions decorator? or in main.py? please verify that

src/fabric_cli/core/fab_interactive.py
```diff
@@ -108,25 +122,40 @@ def start_interactive(self):
...
+
+                except KeyboardInterrupt:
+                    # Handle Ctrl+C gracefully during command input
+                    utils_ui.print("\nUse 'quit' or 'exit' to leave interactive mode.")
+                    continue
+                except Exception as e:
+                    # Handle unexpected errors during prompt processing
+                    utils_ui.print(f"Error in interactive session: {str(e)}")
+                    utils_ui.print("Session will continue. Type 'quit' to exit safely.")
+                    continue
```
ayeshurun: can you explain why we need these changes?

src/fabric_cli/core/fab_interactive.py
aviatco: WDYM?

aviatco: we will get here if for example we pass wrong flag... so we print this - maybe we can skip it

src/fabric_cli/core/fab_interactive.py
aviatco: you right - this can be removed

src/fabric_cli/core/fab_interactive.py
aviatco: today when user hit Ctrl+C  we exit the interactive mode. 
with this change we will exit only on quit/exit same as other platform.

regarding the second exception, these lines handle exceptions in the interactive session loop, they catch errors that occur during: Context retrieval, Prompt rendering, Session management issues
These errors shouldn't terminate the interactive session. The continue statement keeps the session alive, which is the correct behavior for interactive mode (roo :)).

may-hartov: What will happen if user types ctrl+c? It should interrupt the current running command. There are other Python REPL keyboard shortcuts that should terminate the interactive mode - CTRL+D for example, will it work?

src/fabric_cli/main.py
```diff
@@ -82,26 +84,55 @@ def main():
         elif args.version:
             fab_ui.print_version()
         else:
-            # Display help if "fab"
-            fab_ui.display_help(COMMANDS)
+            # AUTO-REPL: When no command is provided, automatically enter interactive mode
```
may-hartov, line 87: Did you consider using a dedicated command? for example, in azure-cli the command is "az interactive". Not sure if this was already discussed and agreed.

src/fabric_cli/main.py
```diff
@@ -56,21 +58,21 @@ def main():
...
-                        subparser = subparsers.choices[command_parts[0]]
-                        subparser_args = subparser.parse_args(command_parts[1:])
-                        subparser_args.command = command_parts[0]
-                        last_exit_code = _execute_command(
-                            subparser_args, subparsers, parser
-                        )
-                        commands_execs += 1
-                        if index != len(args.command) - 1:
-                            fab_ui.print_grey("------------------------------")
+                        if command_parts:  # Ensure we have valid command parts
```
may-hartov, line 61: why is this required? Isn't "if args.command" above sufficient?

src/fabric_cli/main.py
```diff
@@ -82,26 +84,55 @@ def main():
...
+
+def _start_auto_repl():
+    """Start Auto-REPL with proper initialization and error handling."""
+    try:
+        start_interactive_mode()
     except Exception as err:
         fab_ui.print_output_error(
-            FabricCLIError(err.args[0], fab_constant.ERROR_UNEXPECTED_ERROR),
-            output_format_type=args.output_format,
+            FabricCLIError(
```
may-hartov: Can we move it inside start_interactive_mode? just the printing and re-throw. I guess we would like to print it in all places where we call start_interactive_mode

src/fabric_cli/main.py
aviatco: if I understand the flow correctly, here is the case where we have concat commands (in command line mode) so args.command means we gave command provided, command_parts is after we split the commands and here, we check if indeed we have more than one command provided.

src/fabric_cli/main.py
aviatco: the idea was to implement the logic similar to how it is done in node/python etc... for example to start REPL in python you just type python + enter

src/fabric_cli/main.py
aviatco: actually it might be redundant since we have try catch inside start_interactive_mode (in start_interactive).  will check it

src/fabric_cli/core/fab_interactive.py
aviatco: if user types ctrl+c the command will be interrupt but will not terminate the interactive mode - to exit interactive mode user should use quit/exit
I checked ctrl +d - it fails into the error section getting this:
Error in interactive session: 
Session will continue. Type 'quit' to exit safely.

Do you think it is wrong?

may-hartov: we need to check what are the common keyboard shortcuts to exit REPL, I believe ctrl + z and ctrl+ D are exit shortcuts.

src/fabric_cli/main.py
aviatco: removed

aviatco:
@copilot can you find which tests caused the pipeline to canceled?

src/fabric_cli/core/fab_interactive.py
aviatco: I changed the continue to break that also resolve the infinity loop issue in the tests. so now if any errors hits in interactive mode (unless it's ctrl c) we will exit interactive mode

src/fabric_cli/core/fab_interactive.py
may-hartov: so the error will be logged without it, right?

src/fabric_cli/core/fab_interactive.py
ayeshurun: i'm ok with both approaches

PR #111, opened 2026-01-01
refactor(set): refactor the update flow to construct PATCH request bodies correctly 

may-hartov:
**Summary**

Public Fabric APIs for updating resources (capacities, workspaces, items, etc.) are exposed as REST PATCH operations and expect request bodies to include only the fields being changed.
Our CLI set command previously fetched the full resource payload via GET, modified a single field (for example, displayName), and sent the entire object back to the update endpoint. This behavior does not align with PATCH semantics and can result in unexpected behavior or validation errors.
This PR refactors the update flow to construct PATCH request bodies correctly by extracting only the updated properties from the GET response.

**Changes**

- fab_cmd_set_utils:

  1. Extracted update_item_definition from update_fabric_element, as it is item-specific.
  2. Removed json_payload from update_fabric_element return values (unused by callers).
  3. Added extract_updated_only: bool = True to update_fabric_element to control whether only the updated path is extracted or the full payload is returned. For OneLake shortcut updates, this flag is set to False, since shortcuts use the Create API rather than an Update API.
  4. Introduced update_cache to reduce code duplication across fab_fs_set_* implementations.

- fab_fs_set_*
  Removed data.pop(...) calls. Since the payload now includes only the target path being updated, fields like id are no longer present and do not need to be stripped.

- fab_fs_set_connection
  Added connectivityType to the payload, as required by the [Update Gateway API documentation](https://learn.microsoft.com/en-us/rest/api/fabric/core/connections/update-connection).

- fab_fs_set_gateway
  Added the required type field, per the [Update Gateway API documentation.]((https://learn.microsoft.com/en-us/rest/api/fabric/core/connections/update-gateway).

**Tests**
Because this change affects how request payloads are constructed for update APIs, all set command tests were re-recorded to reflect the new behavior.

### DIFF
--- a/src/fabric_cli/commands/fs/set/fab_fs_set_item.py
+++ b/src/fabric_cli/commands/fs/set/fab_fs_set_item.py
@@ -57,7 +57,9 @@ def exec(item: Item, args: Namespace) -> None:
             item_api.update_item(args, item_update_payload, item_uri=True)
 
             if fab_constant.ITEM_QUERY_DISPLAY_NAME in update_payload_dict:
-                new_item_name = update_payload_dict[fab_constant.ITEM_QUERY_DISPLAY_NAME]
+                new_item_name = update_payload_dict[
+                    fab_constant.ITEM_QUERY_DISPLAY_NAME
+                ]
                 item._name = new_item_name
                 utils_mem_store.upsert_item_to_cache(item)
 
@@ -69,7 +71,6 @@ def _update_item_definition(
     query_value: str,
     input_value: str,
 ) -> dict:
-    """Update item definition with base64 decode/encode."""
     try:
         updated_def = utils_set.update_item_definition(
             item_def,
@@ -89,7 +90,6 @@ def _update_item_metadata(
     query_value: str,
     input_value: str,
 ) -> dict:
-    """Update item metadata without base64 encoding."""
     try:
         return utils_set.update_fabric_element(
             item_metadata,
--- a/.changes/unreleased/optimization-20260101-115402.yaml
+++ b/.changes/unreleased/optimization-20260101-115402.yaml
@@ -1,0 +1,3 @@
+kind: optimization
+body: set command - refactor the update flow to construct PATCH request bodies by extracting only the updated properties from the GET payload.
+time: 2026-01-01T11:54:02.344718831Z


src/fabric_cli/commands/fs/set/fab_fs_set_item.py
```diff
@@ -64,29 +57,47 @@ def exec(item: Item, args: Namespace) -> None:
...
+    query_value: str,
+    input_value: str,
+) -> dict:
+    try:
+        return utils_set.update_fabric_element(
+            item_metadata,
+            query_value,
+            input_value,
+        )
+    except (ValueError, KeyError, IndexError):
```
ayeshurun, line 99: maybe use something more generic, like `except Exception:`

src/fabric_cli/commands/fs/set/fab_fs_set_capacity.py
```diff
@@ -29,22 +29,12 @@ def exec(virtual_ws_item: VirtualWorkspaceItem, args: Namespace) -> None:
...
-            data.pop("name", None)
-            data.pop("tags", None)
-            data.pop("fabricId", None)
-            return json.dumps(data, indent=4)
-
-        capacity_update_def = _prep_for_updated_def(updated_def)
-
         utils_ui.print_grey(f"Setting new property for '{virtual_ws_item.name}'...")
-        response = capacity_api.update_capacity(args, capacity_update_def)
+        response = capacity_api.update_capacity(args, json.dumps(updated_def, indent=4))
```
ayeshurun, line 37: why `indent=4` necessary?

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -76,56 +76,87 @@ def update_fabric_element(
     resource_def: dict,
     query: str,
     input: str,
-    decode_encode: bool = False,
-) -> tuple[str, dict]:
+    extract_updated_only: bool = True,
```
ayeshurun, line 79: the name is a bit amigious and unclear. maybe something like 'return_modified_properties' or something like that

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -76,56 +76,87 @@ def update_fabric_element(
...
+    return updated_def
 
-def extract_json_schema(schema: dict, definition: bool = True) -> tuple:
-    name_description_properties = {
-        "displayName": schema.get("displayName", ""),
-        "description": schema.get("description", ""),
-    }
 
-    definition_properties = None
+def update_cache(
```
ayeshurun, line 140: to be generic and used in other places (like sparkpool), consider getting another argument of property name and use it.

something like

```py

def update_catch(
    updated_def: dict,
    element,
    cache_update_func,
    element_name_key = "displayName"
) -> None:
    """Update element's display name in cache if it was changed.

    Args:
        updated_def: Dictionary containing the updated properties.
        element: The Fabric element whose name should be updated.
        cache_update_func: Function to call to update the cache.
        element_name_key: Key of the element name property. Default to 'displayName'.

    """
    if element_name_key in updated_def:
        element._name = updated_def[element_name_key]
        cache_update_func(element)
```

src/fabric_cli/commands/fs/set/fab_fs_set_workspace.py
```diff
@@ -53,11 +46,10 @@ def exec(workspace: Workspace, args: Namespace) -> None:
...
         else:
-            response = workspace_api.update_workspace(args, update_workspace_payload)
+            response = workspace_api.update_workspace(args, json.dumps(updated_def))
 
         if response.status_code == 200:
-            # Update mem_store
-            workspace._name = name_description_properties["displayName"]
-            utils_mem_store.upsert_workspace_to_cache(workspace)
-
+            utils_set.update_cache(
```
ayeshurun, line 52: really liked this refactor!

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -76,56 +76,87 @@ def update_fabric_element(
...
+    try:
+        input = json.loads(input)
+    except (TypeError, json.JSONDecodeError):
+        pass
+
+    # Decode > replace > encode
+    decoded_item_def = _decode_payload(item_def)
+    decoded_item_def = ensure_notebook_dependency(decoded_item_def, query)
+    updated_item_def = utils_jmespath.replace(decoded_item_def, query, input)
+    updated_def = _encode_payload(updated_item_def)
```
ayeshurun, line 135: instead of having this new method, can we reuse the update_fabric_element with another argument like 'update_definition', when true it will do this decode-replace-encode. does that make sense?

src/fabric_cli/commands/fs/set/fab_fs_set_gateway.py
```diff
@@ -35,25 +38,30 @@ def exec(gateway: VirtualWorkspaceItem, args: Namespace) -> None:
...
+            )
+        elif gatewat_type == "OnPremises" and query.startswith(
+            ("numberOfMemberGateways", "capacityId", "inactivityMinutesBeforeSleep")
+        ):
+            raise FabricCLIError(
+                ErrorMessages.Common.gateway_property_not_supported_for_type(
+                    query, "OnPremises"
+                ),
+                fab_constant.ERROR_NOT_SUPPORTED,
+            )
```
ayeshurun, line 56: is it for early failure detection instead of getting api failure?

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: This is very specific to item definitions. For other resources and item metadata, the encode–decode step isn’t needed, so I’d prefer to keep the common flow—used in most places—as simple and clean as possible.

src/fabric_cli/commands/fs/set/fab_fs_set_gateway.py
may-hartov: Each fab_fs_set_* implementation defines an explicit list of supported properties.
For gateways, the supported properties are:
displayName,
allowCloudConnectionRefresh,
allowCustomConnectors,
capacityId,
inactivityMinutesBeforeSleep,
numberOfMemberGateways

The Update Gateway API accepts one of two request body schemas, depending on the gateway type (see [documentation](https://learn.microsoft.com/en-us/rest/api/fabric/core/gateways/update-gateway?tabs=HTTP#request-body)).
If an unsupported property is included in the request body, the API does not fail; the property is simply ignored.
This behavior enforces that update requests conform to a valid schema for the specific gateway type.

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: renamed to return_full_element (inverted)

src/fabric_cli/commands/fs/set/fab_fs_set_capacity.py
may-hartov: it's not. Not sure why it was there in the first place, removed

PR #113, opened 2026-01-01
refactor(set): refactor the update flow to construct PATCH request bodies correctly

may-hartov:
**Summary**

Public Fabric APIs for updating resources (capacities, workspaces, items, etc.) are exposed as REST PATCH operations and expect request bodies to include only the fields being changed.
Our CLI set command previously fetched the full resource payload via GET, modified a single field (for example, displayName), and sent the entire object back to the update endpoint. This behavior does not align with PATCH semantics and can result in unexpected behavior or validation errors.
This PR refactors the update flow to construct PATCH request bodies correctly by extracting only the updated properties from the GET response.

**Changes**

- fab_cmd_set_utils:

  1. Extracted update_item_definition from update_fabric_element, as it is item-specific.
  2. Removed json_payload from update_fabric_element return values (unused by callers).
  3. Added extract_updated_only: bool = True to update_fabric_element to control whether only the updated path is extracted or the full payload is returned. For OneLake shortcut updates, this flag is set to False, since shortcuts use the Create API rather than an Update API.
  4. Introduced update_cache to reduce code duplication across fab_fs_set_* implementations.

- fab_fs_set_*
  Removed data.pop(...) calls. Since the payload now includes only the target path being updated, fields like id are no longer present and do not need to be stripped.

- fab_fs_set_connection
  Added connectivityType to the payload, as required by the [Update Gateway API documentation](https://learn.microsoft.com/en-us/rest/api/fabric/core/connections/update-connection).

- fab_fs_set_gateway
  Added the required type field, per the [Update Gateway API documentation.]((https://learn.microsoft.com/en-us/rest/api/fabric/core/connections/update-gateway).

**Tests**
Because this change affects how request payloads are constructed for update APIs, all set command tests were re-recorded to reflect the new behavior.

### DIFF
--- a/.changes/unreleased/added-20251120-025450.yaml
+++ b/.changes/unreleased/added-20251120-025450.yaml
@@ -1,3 +1,6 @@
 kind: added
 body: Add --format option to export command to pick .ipynb or .py when exporting notebooks
 time: 2025-11-20T02:54:50.543931335Z
+custom:
+  Author: jkafrouni
+  AuthorLink: https://github.com/jkafrouni
--- a/.changes/unreleased/added-20251209-141353.yaml
+++ b/.changes/unreleased/added-20251209-141353.yaml
@@ -1,3 +1,6 @@
 kind: added
-body: Include API response data in the output.
+body: Include API response data in the output
 time: 2025-12-09T14:13:53.030608891Z
+custom:
+  Author: aviatco
+  AuthorLink: https://github.com/aviatco
--- a/.changes/unreleased/added-20251211-183425.yaml
+++ b/.changes/unreleased/added-20251211-183425.yaml
@@ -1,0 +1,6 @@
+kind: added
+body: Display a notification to users on login when a new fab cli version is available
+time: 2025-12-11T18:34:25.601088227+01:00
+custom:
+  Author: Guust-Franssens
+  AuthorLink: https://github.com/Guust-Franssens
--- a/.changes/unreleased/fixed-20251203-155412.yaml
+++ b/.changes/unreleased/fixed-20251203-155412.yaml
@@ -1,3 +1,6 @@
 kind: fixed
-body: Avoid re‑authentication when switching from command‑line to interactive mode.
+body: Avoid reauthentication when switching from command_line to interactive mode
 time: 2025-12-03T15:54:12.961104889Z
+custom:
+  Author: aviatco
+  AuthorLink: https://github.com/aviatco
--- a/.changes/unreleased/fixed-20251228-073108.yaml
+++ b/.changes/unreleased/fixed-20251228-073108.yaml
@@ -1,3 +1,6 @@
 kind: fixed
 body: Fixed set shortcut so it correctly handles target query
 time: 2025-12-28T07:31:08.76557913Z
+custom:
+  Author: may-hartov
+  AuthorLink: https://github.com/may-hartov
--- a/.changes/unreleased/fixed-20260102-172853.yaml
+++ b/.changes/unreleased/fixed-20260102-172853.yaml
@@ -1,0 +1,6 @@
+kind: fixed
+body: Fix a typo in connection roles
+time: 2026-01-02T17:28:53.418628486+01:00
+custom:
+  Author: Guust-Franssens
+  AuthorLink: https://github.com/Guust-Franssens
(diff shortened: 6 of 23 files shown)


src/fabric_cli/commands/fs/set/fab_fs_set_domain.py
```diff
@@ -26,28 +26,18 @@ def exec(virtual_ws_item: VirtualWorkspaceItem, args: Namespace) -> None:
         args.output = None
         vwsi_domain_def = get_domain.exec(virtual_ws_item, args, verbose=False)
 
-        _, updated_def = utils_set.update_fabric_element(
```
aviatco, line 29: what does _ was used for?

src/fabric_cli/commands/fs/set/fab_fs_set_gateway.py
```diff
@@ -35,25 +38,30 @@ def exec(gateway: VirtualWorkspaceItem, args: Namespace) -> None:
...
+            )
+
+        updated_def = utils_set.update_fabric_element(
+            vwsi_gateway_def, query, args.input
         )
 
-        def _prep_for_updated_def(data):
-            data.pop("id", None)
-            # numberOfMemberGateways is supported only for VirtualNetwork type (reason for the whole match statement)
-            match data.get("type"):
```
aviatco, line 45: why did you remove it?

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -76,56 +76,91 @@ def update_fabric_element(
...
     """Update a Fabric resource element using a JMESPath query.
 
     Args:
         resource_def: Resource definition dictionary to modify.
         query: JMESPath expression specifying the path to update.
         input: New value to set.
-        decode_encode: If True, decode/encode base64 payloads. Default False.
+        return_full_element: If True, returns the entire updated resource element.
+            If False, returns only the modified properties extracted by the query path.
+            Defaults to False for backward compatibility.
```
aviatco: Why False for backward compatibility? shouldn't it be True? 
Previous to this change we retuned the entire dict? at least this is what's written in the PR description

tests/test_utils/test_fab_cmd_set_utils.py
```diff
@@ -18,21 +20,53 @@ def test_update_fabric_element_with_json_input_success():
...
 
-    parsed_payload = json.loads(json_payload)
-    assert isinstance(parsed_payload["definition"]["parts"][0]["x"], dict)
-    assert (
-        parsed_payload["definition"]["parts"][0]["x"]["transparency"]["Value"] == "70D"
+
+def test_update_item_definition_with_base64_payload_success():
+    import base64
+
+    # Create a simple item definition with a base64 encoded payload
```
aviatco: please remove unnecessary comments

src/fabric_cli/commands/fs/set/fab_fs_set_domain.py
may-hartov: update_fabric_element previously returned two values, but only one was actually used. Assigning the second value to _ is a Python convention indicating that the value is intentionally ignored. In this PR, the unused return value was removed, and update_fabric_element now returns a single value.

src/fabric_cli/commands/fs/set/fab_fs_set_gateway.py
may-hartov: I moved the validation before _prep_for_updated_def. This function should operate on the payload itself, not validate whether the user input is valid for a given gateway type.

src/fabric_cli/utils/fab_cmd_set_utils.py
may-hartov: Fixed misleading comment

src/fabric_cli/utils/fab_cmd_set_utils.py
```diff
@@ -76,56 +76,91 @@ def update_fabric_element(
...
     """Update a Fabric resource element using a JMESPath query.
 
     Args:
         resource_def: Resource definition dictionary to modify.
         query: JMESPath expression specifying the path to update.
         input: New value to set.
-        decode_encode: If True, decode/encode base64 payloads. Default False.
+        return_full_element: If True, returns the entire updated resource element.
+            If False, returns only the modified properties extracted by the query path.
+            Defaults to True.
```
ayeshurun: default is False :)

ayeshurun (dismissed):

aviatco (dismissed):

ayeshurun (approved):

PR #132, opened 2026-01-12
test: fix failing live tests

v-alexmoraru:
Adds sleep time for tests using mirrored database, null checks & try-except blocks for several tests' corner cases.

### DIFF
--- a/src/fabric_cli/core/fab_exceptions.py
+++ b/src/fabric_cli/core/fab_exceptions.py
@@ -17,7 +17,7 @@ class FabricCLIError(Exception):
         status_code = status_code or DEFAULT_ERROR_CODE
 
         super().__init__(message)
-        self.message = message.rstrip(".") if message else None
+        self.message = message.rstrip(".")
         self.status_code = status_code
 
     @staticmethod
@@ -35,7 +35,7 @@ class FabricCLIError(Exception):
         )
 
     def formatted_message(self, verbose=False):
-        escaped_text = html.escape(self.message) if self.message else ""
+        escaped_text = html.escape(self.message)
 
         return (
             f"[{self.status_code}] {escaped_text}"


src/fabric_cli/core/fab_exceptions.py
```diff
@@ -111,27 +116,32 @@ def __init__(self, response_text):
...
-            else:
-                self.timestamp = None
+        try:
+            response_data = json.loads(response_text) if response_text else {}
+            error_data = response_data.get("error", {})
+            code = error_data.get("code")
+            message = error_data.get("message")
+
+            self.request_id = None
+            self.timestamp = None
```
ayeshurun: you can set those props before parsing the json, and eliminate setting them in 'except' clause.

src/fabric_cli/core/fab_exceptions.py
```diff
@@ -111,27 +116,32 @@ def __init__(self, response_text):
...
+                    message = message.replace(match.group(0), "")
+
+                message = re.sub(r"\n(?=Time:)", "", message)
+                match = re.search(r"Time:(\S+)", message)
+                if match:
+                    self.timestamp = match.group(1)
+                    message = message.replace(match.group(0), "")
+        except (json.JSONDecodeError, TypeError):
+            message = "An error occurred while processing the operation"
+            code = "UnknownError"
```
ayeshurun: to avoid duplications, consider defining them as constants and use them in all places.

src/fabric_cli/core/fab_exceptions.py
```diff
@@ -188,19 +198,24 @@ def __init__(self, response_text):
...
+        try:
+            response_data = json.loads(response_text) if response_text else {}
+            error_data = response_data.get("error", {})
+            code = error_data.get("code")
+            message = error_data.get("message")
+
+            details: list[dict] = error_data.get("details", [])
+
+            # Extract RootActivityId from the details
+            self.request_id = None
```
ayeshurun: same comment as above

tests/test_commands/test_import.py
```diff
@@ -847,6 +848,10 @@ def _import_create_new_item_success(
 ):
     # Setup
     item = item_factory(item_type)
+    
+    if item_type == ItemType.MIRRORED_DATABASE:
+        time.sleep(60)
```
ayeshurun, line 854: can you explain the reason? and isn't 60 second too long?

src/fabric_cli/commands/fs/fab_fs.py
```diff
@@ -301,22 +301,34 @@ def _search_capacity_id(capacity_name: str) -> str | None:
...
-            if response.status_code == 200:
-                return json.loads(response.text)["id"]
-            else:
-                return _search_capacity_id(capacity_name)
-    else:
-        return _search_capacity_id(capacity_name)
+    import time
+    
+    max_retries = 3
+    retry_delay = 2  # seconds
```
ayeshurun: can you please explain the need for retry?

tests/test_commands/test_import.py
v-alexmoraru: There is an API gap when creating a mirrored DB. There are other instances in the codebase where a 60s sleep was employed, so I decided to keep it coherent. I'll leave an explanatory comment too.

src/fabric_cli/core/fab_exceptions.py
```diff
@@ -5,11 +5,15 @@
 import json
 import re
 
+# Constants for common error messages and codes
+DEFAULT_ERROR_MESSAGE = "An error occurred while processing the operation"
+DEFAULT_ERROR_CODE = "UnknownError"
```
ayeshurun: why not taking what's defined in fab_constants.py?

.changes/unreleased/fixed-20260121-141955.yaml
```diff
@@ -0,0 +1,6 @@
+kind: fixed
```
ayeshurun: please delete this file, i've added the skip-changelog label.

src/fabric_cli/core/fab_exceptions.py
v-alexmoraru: Importing them from the constant file would lead to circular imports

src/fabric_cli/core/fab_exceptions.py
```diff
@@ -5,11 +5,15 @@
...
+
 
 class FabricCLIError(Exception):
     def __init__(self, message, status_code=None):
         super().__init__(message)
-        self.message = message.rstrip(".")
+        self.message = message.rstrip(".") if message else None
         self.status_code = status_code
```
ayeshurun: i think it will be beneficial to have default values here, or wrapping the common logic into single private method instead of multiplying the same code

src/fabric_cli/core/fab_exceptions.py
```diff
@@ -5,13 +5,28 @@
...
 class FabricCLIError(Exception):
-    def __init__(self, message, status_code=None):
+    def __init__(self, message=None, status_code=None):
+        # Use default values if not provided
+        message = message or DEFAULT_ERROR_MESSAGE
+        status_code = status_code or DEFAULT_ERROR_CODE
+
         super().__init__(message)
-        self.message = message.rstrip(".")
+        self.message = message.rstrip(".") if message else None
```
ayeshurun: isn't it redundant if we initialize above with default value

PR #142, opened 2026-01-20
test: parametrize tests to cover all supported items + cd recordings

v-alexmoraru:
Several tests performing operations on an item were only testing a specific item, such as Notebook, SQL Database, etc. This PR parametrizes those tests to be run over all the items the operation supports. Some files also include new tests, covering special cases.


tests/test_commands/test_rm.py
ayeshurun, line 1: @v-alexmoraru commenting out should be for new tests, let's avoid from removing/commenting out existing tests..

PR #143, opened 2026-01-20
test(set): Improve set command E2E coverage

may-hartov:
This pull request refactors and enhances the test suite for the `set` command, focusing on improving code reuse, and increasing test coverage for gateway and spark pool metadata updates. 

**Test Refactoring and Parameterization:**

- Refactored `test_set_sparkpool_success` and `test_set_gateway_virtualNetwork_success` to use parameterized test cases, allowing multiple metadata fields and input values to be tested more efficiently. The tests now call a shared helper method (`_test_set_metadata_success`) for improved code reuse and maintainability. [[1]](diffhunk://#diff-1a38c60aaffd61cd671f659c90cd2553a96a600daf77c432fbad1e218819cc9fL269-L309) [[2]](diffhunk://#diff-1a38c60aaffd61cd671f659c90cd2553a96a600daf77c432fbad1e218819cc9fR488-R547)

**Gateway Metadata and Cache Handling:**

- Added parameterized tests for additional gateway metadata fields (`numberOfMemberGateways`, `inactivityMinutesBeforeSleep`, `displayName`) and a new test for updating the `capacityId` field.
- Introduced a new `upsert_gateway_to_cache` fixture to mock gateway cache updates in relevant tests.

### DIFF
--- a/tests/test_commands/test_set.py
+++ b/tests/test_commands/test_set.py
@@ -34,7 +34,7 @@ class TestSET:
         assert_fabric_cli_error,
         mock_questionary_print,
         mock_print_done,
-        upsert_item_to_cache,
+        mock_upsert_item_to_cache,
     ):
         # Setup
         lakehouse = item_factory(ItemType.LAKEHOUSE)
@@ -42,7 +42,7 @@ class TestSET:
         # Reset mocks
         mock_questionary_print.reset_mock()
         mock_print_done.reset_mock()
-        upsert_item_to_cache.reset_mock()
+        mock_upsert_item_to_cache.reset_mock()
 
         # Execute command
         cli_executor.exec_command(
@@ -51,7 +51,7 @@ class TestSET:
 
         # Assert
         assert_fabric_cli_error(constant.ERROR_INVALID_QUERY)
-        upsert_item_to_cache.assert_not_called()
+        mock_upsert_item_to_cache.assert_not_called()
 
     @pytest.mark.parametrize(
         "metadata_to_set,should_upsert_to_cache",
@@ -66,7 +66,7 @@ class TestSET:
         cli_executor,
         mock_questionary_print,
         mock_print_done,
-        upsert_item_to_cache,
+        mock_upsert_item_to_cache,
         metadata_to_set,
         vcr_instance,
         cassette_name,
@@ -76,7 +76,7 @@ class TestSET:
             item_factory(ItemType.NOTEBOOK),
             mock_questionary_print,
             mock_print_done,
-            upsert_item_to_cache,
+            mock_upsert_item_to_cache,
             metadata_to_set,
             cli_executor,
             vcr_instance,
@@ -90,7 +90,7 @@ class TestSET:
         cli_executor,
         mock_questionary_print,
         mock_print_done,
-        upsert_item_to_cache,
+        mock_upsert_item_to_cache,
         workspace,
     ):
         # Setup
@@ -105,7 +105,7 @@ class TestSET:
         # Reset mocks
         mock_questionary_print.reset_mock()
         mock_print_done.reset_mock()
-        upsert_item_to_cache.reset_mock()
+        mock_upsert_item_to_cache.reset_mock()
 
         property_path = (
             "definition.parts[0].payload.datasetReference.byConnection.connectionString"
@@ -120,7 +120,7 @@ class TestSET:
 
         # Assert
         mock_print_done.assert_called_once()
-        upsert_item_to_cache.assert_not_called()
+        mock_upsert_item_to_cache.assert_not_called()
         get(report.full_path, query=property_path)
         assert new_semantic_model_id in str(mock_questionary_print.call_args[0][0])
 
@@ -130,7 +130,7 @@ class TestSET:
         cli_executor,
         mock_questionary_print,
         mock_print_done,
-        upsert_item_to_cache,
+        mock_upsert_item_to_cache,
     ):
         # Setup - Create a new variable library
         variable_library = item_factory(ItemType.VARIABLE_LIBRARY)
@@ -139,7 +139,7 @@ class TestSET:
         # Reset mocks
         mock_questionary_print.reset_mock()
         mock_print_done.reset_mock()
-        upsert_item_to_cache.reset_mock()
+        mock_upsert_item_to_cache.reset_mock()
 
         # Execute command - Set properties with activeValueSetName
         cli_executor.exec_command(
@@ -148,7 +148,7 @@ class TestSET:
 
         # Assert
         mock_print_done.assert_called_once()
-        upsert_item_to_cache.assert_not_called()
+        mock_upsert_item_to_cache.assert_not_called()
 
         # Verify the property was set correctly
         get(variable_library.full_path, query="properties.activeValueSetName")
@@ -162,7 +162,7 @@ class TestSET:
         workspace,
         cli_executor,
         assert_fabric_cli_error,
-        upsert_workspace_to_cache,
+        mock_upsert_workspace_to_cache,
     ):
         # Execute command
         cli_executor.exec_command(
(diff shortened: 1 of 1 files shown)


tests/test_commands/test_set.py
```diff
@@ -812,11 +874,15 @@ def _test_set_metadata_success(
             assert ex.value.status_code in (constant.ERROR_NOT_FOUND, "EntityNotFound")
 
         get(new_entity.full_path, query=metadata_to_set)
-        assert mock_questionary_print.call_args[0][0] == new_metadata_value
+        assert (
+            mock_questionary_print.call_args[0][0].lower() == new_metadata_value.lower()
```
aliabufoul: The original check is more strict.
so just double checking that this change is needed

aliabufoul (dismissed):

tests/test_commands/test_set.py
may-hartov: changed back to case sensitive comparison

tests/test_commands/test_set.py
```diff
@@ -487,38 +516,66 @@ def test_set_connection_success(
...
 
     # region Gateway
+    @pytest.mark.parametrize(
+        "query,input",
+        [
+            ("numberOfMemberGateways", "2"),
+            ("inactivityMinutesBeforeSleep", "60"),
+            ("displayName", None),  # Use None to trigger generate_random_string
+        ],
+    )
```
ohadedry, line 526: Does test_set_gateway_virtualNetwork_success.yaml should be clean due to this change? All new created recordings has different name

tests/test_commands/recordings/test_commands/test_set/test_set_gateway_virtualnetwork_success[capacityId-00000000-0000-0000-0000-000000000000].yaml
```diff
@@ -0,0 +1,504 @@
+interactions:
```
Comment (ohadedry) [inline tests/test_commands/recordings/test_commands/test_set/test_set_gateway_virtualnetwork_success[capacityId-00000000-0000-0000-0000-000000000000].yaml]: Which test created this recording?

tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[name-newName].yaml
```diff
@@ -0,0 +1,821 @@
+interactions:
```
Comment (ohadedry) [inline tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[name-newName].yaml:1]: Which test created this recording? (I only found (name, None) tuple)

tests/test_commands/test_set.py
```diff
@@ -266,47 +266,45 @@ def test_set_sparkpool_invalid_query_failure(
...
-        query,
-        input,
+        metadata_to_set,
+        input_value,
         virtual_item_factory,
         cli_executor,
         mock_questionary_print,
         mock_print_done,
         upsert_spark_pool_to_cache,
+        vcr_instance,
```
aviatco, line 287: there was not recording before?

tests/test_commands/test_set.py
```diff
@@ -266,47 +266,45 @@ def test_set_sparkpool_invalid_query_failure(
         upsert_spark_pool_to_cache.assert_not_called()
 
     @pytest.mark.parametrize(
-        "query, input",
+        "metadata_to_set, input_value",
         [
             ("nodeSize", "Medium"),
```
aviatco, line 271: how do we have 5 parameters but only 2 new recordings?

tests/test_commands/test_set.py
```diff
@@ -450,7 +448,38 @@ def test_set_domain_success(
...
 
     # region Connection
-    def test_set_connection_success(
+    @pytest.mark.parametrize(
+        "metadata_to_set,input_value",
+        [
+            ("privacyLevel", "Organizational"),
+        ],
+    )
+    def test_set_connection_metadata_success(
```
aviatco, line 457: why use parametrize if you have only one value?

tests/test_commands/test_set.py
```diff
@@ -266,47 +266,45 @@ def test_set_sparkpool_invalid_query_failure(
...
 
-        # Reset mocks
-        mock_questionary_print.reset_mock()
-        mock_print_done.reset_mock()
-        upsert_spark_pool_to_cache.reset_mock()
-
-        # Execute command
-        cli_executor.exec_command(
-            f"set {sparkpool.full_path} --query {query} --input {input} --force"
+        self._test_set_metadata_success(
```
aviatco, line 295: I know this isn’t directly related to your PR, but why not use `@pytest.mark.parametrize` for the entity type instead of calling `_test_set_metadata_success` in every test that doesn’t do much before calling it?

tests/test_commands/test_set.py
may-hartov: removed the redundant record file

tests/test_commands/test_set.py
may-hartov: other record files already existed, I added 2 new tests

tests/test_commands/test_set.py
may-hartov: for existing parameters (like autoScale.minNodeCount) - the records exist and should not change. For new parameters ive added the record files

tests/test_commands/test_set.py
may-hartov: It's not really required, just thought is will be easier to extend in the future

tests/test_commands/test_set.py
```diff
@@ -487,38 +516,66 @@ def test_set_connection_success(
     # endregion
 
     # region Gateway
+    @pytest.mark.parametrize(
+        "query,input",
```
aviatco, line 520: Let’s stay consistent and call it `metadata_to_set` and `input_value`, since you already changed it in other tests.

tests/test_commands/test_set.py
```diff
@@ -918,6 +982,18 @@ def upsert_folder_to_cache():
         yield mock
 
 
+@pytest.fixture()
+def upsert_gateway_to_cache():
```
aviatco: consider add prefix mock_ so it will be clear this is a mock for upsert_gateway_to_cache

tests/test_commands/test_set.py
may-hartov: We can do it, then we will have to parametrize also the injected factory (for some we use workspace_factory for some virtual_workspace_item_factory).... I will consider in a separate PR

tests/test_commands/recordings/test_commands/test_set/test_set_gateway_virtualnetwork_success[capacityId-00000000-0000-0000-0000-000000000000].yaml
Comment (may-hartov) [reply tests/test_commands/recordings/test_commands/test_set/test_set_gateway_virtualnetwork_success[capacityId-00000000-0000-0000-0000-000000000000].yaml]: deleted

tests/test_commands/test_set.py
aviatco: where is it used?

tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[name-newName].yaml
Comment (may-hartov) [reply tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[name-newName].yaml:1]: deleted

tests/test_commands/test_set.py
```diff
@@ -918,6 +982,18 @@ def upsert_folder_to_cache():
...
+    with patch("fabric_cli.utils.fab_mem_store.upsert_gateway_to_cache") as mock:
+        yield mock
+
+
+@pytest.fixture()
+def upsert_connection_to_cache():
```
aviatco: consider add prefix mock_ so it will be clear this is a mock

tests/test_commands/test_set.py
aviatco: my question was why we only now added  vcr_instance, and  cassette_name? this wasn't needed before?

may-hartov: We need to inject it to _test_set_metadata_success which we didnt use before

tests/test_commands/test_set.py
```diff
@@ -817,6 +879,8 @@ def _test_set_metadata_success(
         # Clean up - update the full path of the renamed entities so the factory can clean them up
         if metadata_to_set == "displayName":
             set(new_entity.full_path, query="displayName", input=entity.display_name)
+        elif metadata_to_set == "name":
+            set(new_entity.full_path, query="name", input=entity.display_name)
```
ohadedry, line 883: if query is name, does input is entity.display_name?

may-hartov: Yes. EntitiyMetadata has the Fabric resource type suffix - some-name.Type while displayName doesnt have the suffix.

PR #152, opened 2026-01-30
test: add recordings for new assign tests

v-alexmoraru:


tests/test_commands/test_assign.py
```diff
@@ -13,6 +15,65 @@
 
 
 class TestAssign:
+    # region Parametrized Tests
+    def test_assign_domain_workspace_success(
```
ayeshurun: @v-alexmoraru can this test be parameterized?

tests/test_commands/test_assign.py
```diff
@@ -13,6 +15,65 @@
...
+        get(domain.full_path, query="domainWorkspaces")
+        assert any(
+            workspace.display_name in str(call.args[0])
+            for call in mock_questionary_print.mock_calls
+        )
+
+    @pytest.mark.parametrize("entity_type,factory_key,path_template", [
+        (VirtualWorkspaceType.CAPACITY, "test_data", "/.capacities/{}.Capacity"),
+        (VirtualWorkspaceType.DOMAIN, "virtual_workspace_item_factory", "{}.full_path"),
+    ])
```
ayeshurun: @v-alexmoraru can this placed in confttest?

tests/test_commands/test_assign.py
v-alexmoraru: I don't think so, I moved it from the parametrized section

PR #155, opened 2026-01-30
test: add recordings for new export tests

v-alexmoraru:


tests/test_commands/test_export.py
```diff
@@ -64,31 +88,43 @@ def test_export_item_home_directory_path_success(
...
+        ItemType.SPARK_JOB_DEFINITION,
+        ItemType.DATA_PIPELINE,
+        ItemType.MIRRORED_DATABASE,
+        ItemType.REPORT,
+        ItemType.SEMANTIC_MODEL,
+        ItemType.KQL_DATABASE,
+        ItemType.COSMOS_DB_DATABASE,
+        ItemType.USER_DATA_FUNCTION,
+        ItemType.GRAPH_QUERY_SET
+    ])
```
ayeshurun: how can we maintain one list instead of defining anytime new mark.parametrize ? i am trying to think how we will maintain it going forward

tests/test_commands/test_export.py
```diff
@@ -4,42 +4,64 @@
...
 
+    @pytest.mark.parametrize("item_type,expected_file_extension", [
+        (ItemType.NOTEBOOK, ".ipynb"),
+        (ItemType.SPARK_JOB_DEFINITION, ".json"),
+        (ItemType.DATA_PIPELINE, ".json"),
+        (ItemType.MIRRORED_DATABASE, ".json"),
+        (ItemType.COSMOS_DB_DATABASE, ".json"),
+        (ItemType.USER_DATA_FUNCTION, ".json"),
+        (ItemType.GRAPH_QUERY_SET, ".json")
+    ])
```
ayeshurun: is this the same as above ? if yes, let's define in conftest and use it here and above.

btw, i think it will be best to concentrate all markers like this one in conftest, even if they are too specific for the scenario - perhaps we can name them according to the scenario they are being used. this way we will have one place to maintain and strightforward decision what items i want to add / update.

ayeshurun (approved):

PR #157, opened 2026-01-30
test: add recordings for new import tests

v-alexmoraru:


tests/test_commands/test_import.py
```diff
@@ -15,532 +15,54 @@
...
-            cli_executor,
-        )
+    @pytest.fixture(scope="class", autouse=True)
+    def reset_counter(self):
+        cassette_resource_counters.clear()
+        yield
 
-    def test_import_create_new_reflex_item_success(
+    # region Parametrized Tests
+    @import_update_existing_item_success_params
```
ohadedry, line 32: If this replaces all specific test per item, for example this replace test_import_update_existing_notebook_item_success, then we don't need the recording test_import_update_existing_notebook_item_success.yaml, right?
(I think the new recording is test_import_update_existing_item_success[Notebook].yaml)

Same for other items.

tests/test_commands/test_import.py
```diff
@@ -15,665 +15,55 @@
...
-            spy_create_item,
-            mock_print_grey,
-            upsert_item_to_cache,
-            ItemType.MIRRORED_DATABASE,
-            cli_executor,
-        )
+    @pytest.fixture(scope="class", autouse=True)
+    def reset_counter(self):
+        cassette_resource_counters.clear()
+        yield
```
ayeshurun: @v-alexmoraru can you please explain this change?

v-alexmoraru: Leftover code from an older approach. Deleted it

PR #159, opened 2026-01-30
test: add recordings for new mkdir tests

v-alexmoraru:


tests/test_commands/test_mkdir.py
```diff
@@ -206,6 +242,104 @@ def test_mkdir_kqldatabase_without_creation_payload_success(
...
+        assert item_display_name in result_output
+
+        # Check type-specific assertions
+        for assertion in expected_assertions:
+            assert assertion in result_output
+
+        # Cleanup
+        rm(item_full_path)
+
+    def test_mkdir_mounted_data_factory_with_required_params_success(
```
ayeshurun, line 289: @v-alexmoraru why separate test for this item? can't we leverage the parameterize?

v-alexmoraru: It requires a bit of a different approach, with parameters such as `subscriptionId`, or `resourceGroup`

tests/test_commands/test_mkdir.py
```diff
@@ -206,6 +242,104 @@ def test_mkdir_kqldatabase_without_creation_payload_success(
...
+        # Verify item was created
+        mock_questionary_print.reset_mock()
+        get(mdf_full_path, query=".")
+        mock_questionary_print.assert_called_once()
+        assert mdf_display_name in mock_questionary_print.call_args[0][0]
+
+        # Cleanup
+        rm(mdf_full_path)
+
+    def test_mkdir_mounted_data_factory_missing_required_params_failure(
```
ayeshurun, line 323: @v-alexmoraru also here

v-alexmoraru: Are there other item types that require mandatory params? If not, this should suffice

ayeshurun (approved):

PR #160, opened 2026-01-30
test: add recordings for new unassign tests

v-alexmoraru:


tests/test_commands/conftest.py
```diff
@@ -56,6 +56,18 @@
     "x-ms-authorization-auxiliary",
 ]
 
+unassign_entity_workspace_success_params = pytest.mark.parametrize("entity_type,factory_key,path_template,assertion_key", [
+    (VirtualWorkspaceType.CAPACITY, "test_data",
+        "/.capacities/{}.Capacity", "id"),
+    (VirtualWorkspaceType.DOMAIN, "virtual_workspace_item_factory",
+        "{}.full_path", "display_name"),
```
ohadedry: According to test_unassign.py `path_template` is not used when factory_key is not test_data. I think it is better to use `None` instead of `{}.full_path` since it is misleading that the test actually using it. 
What about using `full_path` as param and in test use `getattr(item, path_template, None)`

tests/test_commands/test_unassign.py
```diff
@@ -4,16 +4,127 @@
...
 from fabric_cli.core import fab_handle_context as handle_context
 from fabric_cli.core.fab_types import ItemType, VirtualWorkspaceType
 from tests.test_commands.data.static_test_data import StaticTestData
+from tests.test_commands.conftest import unassign_entity_workspace_success_params, unassign_failure_params
 
 
 class TestUnassign:
+    # region Parametrized Tests
+    @unassign_entity_workspace_success_params
+    def test_unassign_entity_workspace_success(
```
ohadedry: test has too many if/else to differentiate between Capacity and Domain which are the only cases. What was the reason for using here Parametrized tests?

v-alexmoraru: I think I initially had several params that I ended up removing. I'll split the test in 2 unparametrized ones.

PR #165, opened 2026-02-03
style: add visual divider between folders and items in workspace listing

ayeshurun:
# 📥 Pull Request

This PR enhances the `fab ls` command by adding a visual divider to improve the readability of workspace listings.

- **Visual Divider**: A `------------------------------` divider now separates folders from items in the output.
- **Conditional Display**: The divider only appears when `folder_listing_enabled=true`, `output_format=text`, and the `--long` flag is **not** used.
- **Implementation**: The divider is added as a data entry, avoiding changes to core rendering logic.

This change provides a clearer distinction between content types in the workspace, improving user experience for large workspaces.

### Example Output

```
Sales.Folder
Marketing.Folder
Engineering.Folder
------------------------------
Product-Catalog.SemanticModel
Sales-Dashboard.Report
Inventory.SemanticModel
Customer-Data.Report
```


src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
```diff
@@ -10,12 +10,68 @@
...
+    return result
+
 
 def exec(workspace: Workspace, args):
     show_details = bool(args.long)
     show_all = bool(args.all)
     ws_elements: list[Union[Item, Folder]] = utils_fs.get_ws_elements(workspace)
-    sorted_elements_dict = utils_fs.sort_ws_elements(ws_elements, show_details)
+
+    # Check if folder listing is enabled
```
aviatco: pls remove redundant comment - same for other comments bellow

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
```diff
@@ -10,12 +10,68 @@
...
+    folder_listing_enabled = (
+        fab_state_config.get_config(fab_constant.FAB_FOLDER_LISTING_ENABLED) == "true"
+    )
+
+    # Get output format from args or config
+    output_format = getattr(args, "output_format", None) or fab_state_config.get_config(
```
aviatco: output_format should not be used here - only in fab_ui

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
```diff
@@ -10,12 +10,68 @@
...
+    first_group = True
+    type_order = [Folder, Item]
+
+    for typ in type_order:
+        group = [el for el in ws_elements if isinstance(el, typ)]
+        if group:
+            group_dicts = utils_fs.sort_ws_elements(group, show_details)
+            if not first_group:
+                divider = {"name": _DIVIDER}
+                if show_details:
```
aviatco: didn't show_details is always false?

 if folder_listing_enabled and output_format == "text" and not show_details:
        sorted_elements_dict = _sort_ws_elements_with_seperation(
            ws_elements, show_details
        )

tests/test_commands/test_ls.py
```diff
@@ -1203,6 +1203,69 @@ def test_ls_folder_content_success(
...
+        cli_executor: CLIExecutor,
+    ):
+        """Test that a divider appears between folders and items when folder_listing_enabled=true"""
+        # Setup: Create folders and items in workspace root
+        folder1 = folder_factory()
+        folder2 = folder_factory()
+        notebook = item_factory(ItemType.NOTEBOOK, path=workspace.full_path)
+        lakehouse = item_factory(ItemType.LAKEHOUSE, path=workspace.full_path)
+
+        # Test 1: List workspace (folder_listing_enabled is true by default in tests)
```
aviatco: pls remove redundant comments

tests/test_commands/test_ls.py
```diff
@@ -1203,6 +1203,69 @@ def test_ls_folder_content_success(
...
+        cli_executor.exec_command(f"ls {workspace.full_path}")
+
+        # Assert: Check that the divider is present
+        mock_questionary_print.assert_called()
+
+        # Verify divider appears in output
+        _assert_strings_in_mock_calls(
```
aviatco: do we want to verify the order it correct? folders ->  separator -> items ?

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
```diff
@@ -10,12 +10,68 @@
 from fabric_cli.utils import fab_cmd_fs_utils as utils_fs
 from fabric_cli.utils import fab_cmd_ls_utils as utils_ls
 
+# Divider used to separate folders and items in workspace listings
+_DIVIDER = "------------------------------"
+
+
+def _sort_ws_elements_with_seperation(
```
aviatco: why not adding the separator in get_ws_elements? we already filter there the folders and items... so we could add there the login

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
```diff
@@ -10,12 +10,68 @@
 from fabric_cli.utils import fab_cmd_fs_utils as utils_fs
 from fabric_cli.utils import fab_cmd_ls_utils as utils_ls
 
+# Divider used to separate folders and items in workspace listings
+_DIVIDER = "------------------------------"
```
aviatco: we have this logic in fab_ui when hidden_data is true - why not enhance this logic? it will resolve also the use of output_format in ls command

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
ayeshurun: Done — removed all redundant comments from the command file.

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
ayeshurun: Done — removed the `output_format` check from the command. The divider logic now applies regardless of output format.

tests/test_commands/test_ls.py
ayeshurun: Done — cleaned up redundant comments in the tests.

tests/test_commands/test_ls.py
ayeshurun: Added ordering assertions — the test now verifies folders appear before the `[Items]` header and items appear after it.

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
ayeshurun: Good catch — removed the `show_details` condition. `sort_ws_elements_with_divider()` now works for both `--long` and default modes, inserting `[Folders]`/`[Items]` group headers when both exist.

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
ayeshurun: Moved the divider logic to a dedicated `sort_ws_elements_with_divider()` utility function in `fab_cmd_fs_utils.py`. This keeps `get_ws_elements()` focused on data retrieval while the new function handles sorting + group header insertion. The command file just calls it when `folder_listing_enabled` is true.

src/fabric_cli/commands/fs/ls/fab_fs_ls_item.py
ayeshurun: Considered this, but `hidden_data` is already used for `VirtualItemContainerType` (the `--all/-a` flag renders `.sparkpools`, `.managedidentities`, etc. below the divider). Using it for folder/item grouping would conflict with that existing behavior. Instead, we insert `[Folders]`/`[Items]` group header dicts directly in the data list via `sort_ws_elements_with_divider()`, which works cleanly with both text and `--long` output modes while preserving the existing `hidden_data` functionality.

PR #169, opened 2026-02-05
fix: fix support of cosmos db in ls command

ayeshurun:


tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml
```diff
@@ -17,7 +17,7 @@ interactions:
   response:
```
Comment (ohadedry) [inline tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml]: Why recording this test is relevant to the fix?

tests/test_commands/recordings/test_commands/test_ls/test_ls_item_folders_success[Lakehouse-expected_folders0].yaml
```diff
@@ -0,0 +1,665 @@
+interactions:
```
Comment (ohadedry) [inline tests/test_commands/recordings/test_commands/test_ls/test_ls_item_folders_success[Lakehouse-expected_folders0].yaml:1]: If you changed the original test, then the previous recording is redundant test_ls_item_folders_success.yaml

tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml
Comment (ohadedry) [inline tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml:1]: Why did you update this test and the class_setup?

ohadedry (commented):
The PR is about ls command but the item type is also supported for other commands

tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml
Comment (ayeshurun) [reply tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml:1]: the test failed locally so i re-recorded.

tests/test_commands/recordings/test_commands/test_ls/test_ls_item_folders_success[Lakehouse-expected_folders0].yaml
Comment (ayeshurun) [reply tests/test_commands/recordings/test_commands/test_ls/test_ls_item_folders_success[Lakehouse-expected_folders0].yaml:1]: Good catch. removed

PR #170, opened 2026-02-08
fix: fix open dtb url

ayeshurun:


tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml
```diff
@@ -1,670 +1,684 @@
 interactions:
```
Comment (may-hartov) [inline tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml:1]: why did you need to change tohis record?

Comment (ayeshurun) [reply tests/test_commands/recordings/test_commands/test_set/test_set_sparkpool_success[autoScale.enabled-True].yaml:1]: Reformat changes

PR #190, opened 2026-03-17
perf: lazy imports and session reuse for 93% faster CLI startup

ayeshurun:
# 📥 Pull Request

Deferred heavy imports (`msal`, `jwt`, `cryptography`, `requests`, `prompt_toolkit`, `psutil`) that were loaded on every CLI invocation — even for `fab --version`.

### Changes

- **Lazy command loading** — new `lazy_command()` utility in `fab_lazy_load.py` defers command module imports in all 9 parser files until the command is actually invoked
- **Deferred auth & interactive imports** in `main.py` — auth stack (~160ms) and prompt_toolkit (~64ms) only load when needed
- **Deferred `psutil`** in `fab_context.py` — moved into `_get_context_session_id()` with `TYPE_CHECKING` guard for mypy
- **Config write optimization** in `fab_state_config.py` — `init_defaults()` skips file write when nothing changed
- **HTTP session reuse** in `fab_api_client.py` — shared `requests.Session` with retry adapter for connection pooling
- **Extracted** `fab_describe_commands.py` — separated heavy describe logic from parser registration

### Results

| Metric | Before | After |
|--------|--------|-------|
| Module import time | ~737ms | ~54ms |
| Heavy deps at startup | msal, jwt, cryptography, requests, prompt_toolkit | None |

8 new tests covering lazy loading, startup threshold, session reuse, and config optimization.


tests/test_utils/test_lazy_load.py
```diff
@@ -0,0 +1,186 @@
...
+        # Ensure the module is not loaded
+        was_loaded = mod_path in sys.modules
+
+        wrapper = lazy_command(mod_path, "init")
+
+        # The wrapper should be callable
+        assert callable(wrapper)
+
+        # If the module wasn't loaded before, it should still not be loaded
+        if not was_loaded:
```
ohadedry: I didn't understand this if, do we expect that the module is loaded?

tests/test_utils/test_lazy_load.py
```diff
@@ -0,0 +1,186 @@
...
+        from fabric_cli.utils.fab_lazy_load import lazy_command
+
+        # Create a test module with a mock function
+        import types
+
+        test_mod = types.ModuleType("_test_lazy_mod")
+        test_mod.my_func = MagicMock(return_value=42)
+        sys.modules["_test_lazy_mod"] = test_mod
+
+        try:
```
ohadedry, line 52: why do we need try? errors are not handled with `except`

tests/test_utils/test_lazy_load.py
```diff
@@ -0,0 +1,186 @@
...
+                )
+        finally:
+            # Restore all modules
+            for mod_name, mod in saved_modules.items():
+                sys.modules[mod_name] = mod
+            for mod_name, mod in saved_heavy.items():
+                sys.modules[mod_name] = mod
+
+
+class TestSessionReuse:
```
ohadedry: this is not related to lazy load

src/fabric_cli/commands/desc/fab_describe.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) Microsoft Corporation.
```
ohadedry, line 1: commands should be under commands folder, not parsers folder

src/fabric_cli/core/fab_context.py
```diff
@@ -293,6 +296,8 @@ def _get_context_session_id(self):
         """
         parent_process = None
         try:
+            import psutil
```
ohadedry: In continue to what we discussed, I'm not sure what this change saves since `_get_context_session_id` is called from the constructor of context

src/fabric_cli/commands/desc/fab_describe.py
```diff
@@ -0,0 +1,152 @@
...
+# Licensed under the MIT License.
```
aviatco, line 2: is it related to the PR?

src/fabric_cli/main.py
```diff
@@ -31,19 +29,27 @@ def main():
...
                     fab_state_config.get_config(fab_constant.FAB_MODE)
                     == fab_constant.FAB_MODE_INTERACTIVE
                 ):
+                    from fabric_cli.core.fab_interactive import start_interactive_mode
+
                     start_interactive_mode()
                     return
 
         if args.command == "auth" and args.auth_command == "logout":
+            from fabric_cli.commands.auth import fab_auth as login
```
aviatco: nit - if you already move the import inside consider remove the alias and call it fab_auth cause it sounds wired login.logout....

tests/test_utils/test_lazy_load.py
aviatco: I believe it's for the finally

tests/test_utils/test_lazy_load.py
```diff
@@ -0,0 +1,186 @@
...
+        assert session1 is session2, "Session should be reused"
+
+        # Clean up
+        fab_api_client._shared_session = None
+
+
+class TestInitDefaults:
+    """Test suite for config initialization optimization."""
+
+    def test_init_defaults__no_write_when_unchanged(self, tmp_path, monkeypatch):
```
aviatco: double __ in the name

tests/test_utils/test_lazy_load.py
```diff
@@ -0,0 +1,186 @@
...
+        session2 = fab_api_client._get_session()
+
+        assert session1 is session2, "Session should be reused"
+
+        # Clean up
+        fab_api_client._shared_session = None
+
+
+class TestInitDefaults:
+    """Test suite for config initialization optimization."""
```
aviatco: Also this is not related to lazy load - maybe inder: test_config_initialization?

src/fabric_cli/client/fab_api_client.py
```diff
@@ -29,20 +29,39 @@
...
-    timeout_sec=240,
+    timeout_sec=None,
     continuation_token=None,
     hostname=None,
 ) -> ApiResponse:
     json_file = getattr(args, "json_file", None)
     audience_value = getattr(args, "audience", None)
     headers_value = getattr(args, "headers", None)
     method = getattr(args, "method", "get")
+    if timeout_sec is None:
```
aviatco: this is timeout for polling? are we sure that 30 is enough? get item definition can take some time - never counted :)

src/fabric_cli/core/fab_context.py
ayeshurun: You're right — `fab_context` is imported by essentially every command path, so psutil loads regardless. Reverted to top-level `import psutil` matching main.

src/fabric_cli/commands/desc/fab_describe.py
ayeshurun: Moved to `commands/desc/fab_describe.py`.

tests/test_utils/test_lazy_load.py
ayeshurun: Intentional — follows the project's test naming convention `test_<behavior>__<condition>` (per `test.instructions.md`). The double underscore separates what's being tested from the scenario.

tests/test_utils/test_lazy_load.py
ayeshurun: As @aviatco noted — the `try` is paired with `finally` to ensure the test module is cleaned up from `sys.modules` even when assertions fail. No `except` needed since we want exceptions to propagate normally.

tests/test_utils/test_lazy_load.py
ayeshurun: Agreed — moved to `tests/test_utils/test_fab_state_config.py` (alongside @ohadedry's same feedback on `TestSessionReuse`).

tests/test_utils/test_lazy_load.py
ayeshurun: Agreed — moved `TestSessionReuse` to `tests/test_utils/test_fab_api_client.py` and `TestInitDefaults` to `tests/test_utils/test_fab_state_config.py`.

tests/test_utils/test_lazy_load.py
ayeshurun: Good catch — the guard was defensive for test ordering, but it weakened the assertion. Fixed: the test now pops the module from `sys.modules` before running, so the assertion always executes.

src/fabric_cli/client/fab_api_client.py
ayeshurun: Good call — reverted to `timeout_sec=240` for all methods, matching the original default. With 65+ callers all relying on the default and no explicit overrides, reducing GETs to 30s risked regressions on slower list/read operations. The session reuse + retry adapter already provide the perf gains here.

src/fabric_cli/commands/desc/fab_describe.py
ayeshurun: Yes — this extraction is part of the lazy-loading optimization. Previously, the describe logic was inline in the parser module, causing all its heavy imports (hierarchy types, command registry) to load during parser registration at startup. Extracting it allows deferring those imports via `lazy_command()` until `desc` is actually invoked. Per @ohadedry's feedback, also moved it from `parsers/` to `commands/desc/fab_describe.py`.

src/fabric_cli/main.py
ayeshurun: Done — dropped the alias, using `fab_auth` directly. `fab_auth.logout(args)` reads much better.

PR #191, opened 2026-03-17
feat(types): add support for Map item type

ayeshurun:
# 📥 Pull Request

Adds the `Map` item type as a new simple Fabric item (no definition formats, no jobs, no OneLake folders, no creation parameters).

- **Summary**: Onboards the `Map` item type across the CLI codebase and test infrastructure.
- **Context**: The `Map` item type needs to be recognized by the CLI for standard operations like `ls`, `get`, `rm`, `mkdir`, etc.
- **Dependencies**: None.

### Changes

- **`src/fabric_cli/core/fab_types.py`**: Registered `MAP = "Map"` enum member, added `"maps"` to both `format_mapping` (API URI) and `uri_mapping` (portal URI).
- **`src/fabric_cli/core/hiearchy/fab_item.py`**: Added `ItemType.MAP` to the standard multi-case match in `get_payload()` for import support.
- **`tests/test_commands/conftest.py`**: Added `ItemType.MAP` to `ALL_ITEM_TYPES` and `basic_item_parametrize` lists so the new type is covered by all parameterized command tests.

---

💬 Send tasks to Copilot coding agent from [Slack](https://gh.io/cca-slack-docs) and [Teams](https://gh.io/cca-teams-docs) to turn conversations into code. Copilot posts an update in your thread when it's finished.

### DIFF
--- a/tests/test_commands/conftest.py
+++ b/tests/test_commands/conftest.py
@@ -91,7 +91,7 @@ mv_item_to_item_success_params = pytest.mark.parametrize("item_type", [
     ItemType.DATA_PIPELINE, ItemType.KQL_DASHBOARD, ItemType.KQL_QUERYSET,
     ItemType.MIRRORED_DATABASE, ItemType.NOTEBOOK,
     ItemType.REFLEX, ItemType.SPARK_JOB_DEFINITION,
-    ItemType.COSMOS_DB_DATABASE, ItemType.USER_DATA_FUNCTION, ItemType.GRAPH_QUERY_SET, ItemType.MAP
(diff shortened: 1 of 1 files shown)


src/fabric_cli/core/fab_config/command_support.yaml
```diff
@@ -199,6 +200,7 @@ commands:
           - dataflow
           - sql_database
           - user_data_function
+          - cp
```
aviatco: cp? did you mean map?

ayeshurun:
```suggestion
          - map
```

ayeshurun: yes my bad. fixed

PR #212, opened 2026-04-15
feat: add support for environment definitions

v-alexmoraru:
Removes the environment workarounds and records new tests for the enabled commands.


src/fabric_cli/commands/fs/impor/fab_fs_import_item.py
```diff
@@ -53,11 +52,7 @@ def import_single_item(item: Item, args: Namespace) -> None:
                     f"Importing (update) '{_input_path}' → '{item.path}'..."
                 )
 
-                # Environment item type, not supporting definition yet
-                if item.item_type == ItemType.ENVIRONMENT:
-                    _import_update_environment_item(args, payload)
-                else:
-                    _import_update_item(args, payload)
+                _import_update_item(args, payload)
```
ayeshurun: What are the implications of moving to the definitions approach? with the current approach we are doing some manual stuff, and at the end publishing the environment. 

Did we compare the end result of two approaches? Perhaps we should keep support somehow of these  two

src/fabric_cli/commands/fs/impor/fab_fs_import_item.py
```diff
@@ -66,11 +61,7 @@ def import_single_item(item: Item, args: Namespace) -> None:
             utils_ui.print_grey(
                 f"Importing '{_input_path}' → '{item.path}'...")
 
-            # Environment item type, not supporting definition yet
-            if item.item_type == ItemType.ENVIRONMENT:
-                response = _import_create_environment_item(item, args, payload)
-            else:
-                response = _import_create_item(args, payload)
+            response = _import_create_item(args, payload)
```
ayeshurun: same question as above.

src/fabric_cli/commands/fs/impor/fab_fs_import_item.py
ayeshurun: just want to make sure that current version available today vs your changes produce the same results and state of item.

ayeshurun (approved):

PR #216, opened 2026-04-17
fix: wrap long text on multiple rows

### LINKED ISSUE #201: [BUG] Table output columns break when content exceeds terminal width
## Description

When `print_output_format` renders a table with `show_headers=True` and the total row width exceeds the terminal width, columns are pushed off-screen and wrap to the next line. This makes the output appear as if columns have disappeared: headers show on one line, some values appear on the next line, and the table alignment is completely lost.

### fabric-cli version

1.3.1

### Python version

Python 3.12.10

### Operating System

Windows

### CLI Mode

Interactive mode

### Authentication Method

Interactive browser login

### Steps to Reproduce

1. Use a command like ls or find that supports table rendering
2. Get a response with long values

### Expected Behavior

Table columns should fit within the terminal width. Long values in columns should be truncated with `...` to prevent the table from exceeding the terminal width.

### Actual Behavior

When a page contains column with long values, the table row exceeds the terminal width. This causes:
- Headers are pushed to the far right edge or off-screen
- Values to wrap to a second line
- The table to appear as if columns are missing

### Additional Context

This affects all commands that use `print_output_format` with `show_headers=True`.

### Possible Solution

- The `find` command already truncates `description` in compact mode using `shutil.get_terminal_size()`. A similar approach could be applied to other columns like `name`.
- A more robust fix would be terminal-width-aware column sizing in the shared table renderer (`_print_output_format_result_text`).

v-alexmoraru:
- Fixed #201 

When `print_output_format` rendered a table with `show_headers=True` and long column values, rows exceeded the terminal width causing headers and values to appear misaligned or off-screen.

- Added `_wrap_text` helper that splits text into lines respecting visual width
- Updated `_format_unix_style_field` and `_format_unix_style_entry` to return list[str] — one line per wrapped row — instead of a single potentially overflowing string
- Updated `print_entries_unix_style` to cap column widths proportionally to fit the terminal and print each new row


src/fabric_cli/utils/fab_ui.py
```diff
@@ -288,7 +293,13 @@ def print_entries_unix_style(
 
     if header:
         widths = [
-            max(len(field), max(get_visual_length(entry, field) for entry in _entries))
+            max(
```
aviatco, line 296: there is a new function recently added - fab_utils.truncate_columns maybe you can try reuse it to fix the issue

v-alexmoraru: The existing truncate method hides the overflowing text behind `...`. The approach I used adds it on a new line below, so that no text is hidden. I think it's cleaner to keep it this way

aviatco: I see, but it this case we will have 2 different options for displaying the text. so I think we need to decide which one we select and remove the other.
Also, the idea behind the ... was to not have long text in the tables since it doesn't look good.. so adding "..." keep it short and clean.

aviatco: can you share how does it looks with your fix and how with "..."

v-alexmoraru: Multiple rows:
multiple_rows

Trimmed:
trimmed

aviatco: with the trim option, if we hover on the "..." we see the rest of the text? or how can we see it?
and in the Multiple rows, who define the size of the row? why the first and last result are shorter than the middle?

v-alexmoraru: With the trimming approach, we can't see the text covered by "...". I'm not sure if hovering is something achievable in our case, although it would be nice.

Regarding the several row approach, the size of the row is defined by the size of the terminal. The executions have different sizes intentionally - I resized the terminal to showcase how the text would behave in those circumstances. So every execution is depending on the terminal size in the moment we run the command.

aviatco (dismissed):

src/fabric_cli/utils/fab_ui.py
aviatco: ok so lets use your approach. I approve the PR. just fix the conflicts so I can merge it

aviatco (approved):

PR #233, opened 2026-05-07
refactor: show error message

v-alexmoraru:
Shows Fabric API error message more explicitly.


src/fabric_cli/core/fab_exceptions.py
```diff
@@ -78,12 +82,23 @@ def __init__(self, response_text):
...
         """
-        response = self._parse_json_response(response_text)
-
-        message = response.get("message")
-        error_code = response.get("errorCode")
-        self.more_details: list[dict] = response.get("moreDetails", [])
-        self.request_id = response.get("requestId")
+        try:
+            response = json.loads(response_text)
+            if not isinstance(response, dict):
```
aviatco, line 87: I think response_text can be a string, not always an object. but Im not sure. worth checking it

v-alexmoraru: In that case it's caught and returned by `except` block

PR #235, opened 2026-05-13
feat: supports digital twin builder flow item

v-alexmoraru:
Onboards Digital Twin Builder Flow item type & records tests for its commands.


src/fabric_cli/utils/fab_cmd_mkdir_utils.py
```diff
@@ -89,6 +89,41 @@ def add_type_specific_payload(item: Item, args, payload):
                     "parentEventhouseItemId": _eventhouse_id,
                 }
 
+        case ItemType.DIGITAL_TWIN_BUILDER_FLOW:
+            _digital_twin_builder_id = params.get("digitaltwinbuilderid")
+            _workspace_id = item.workspace.id
+
+            if _digital_twin_builder_id:
+                payload_dict["creationPayload"] = {
```
aviatco: since both conditions adding the same creationMethod to payload, consider changing the check to
if !digital_twin_builder_id:
     create ......

then no need for else, just set the payload_dict["creationPayload"] = ......

tests/test_commands/conftest.py
```diff
@@ -44,136 +44,263 @@
 
 ALL_ITEM_TYPES = [
```
aviatco, line 45: it is very hard to review this page because of all the changes... can you explain what you did? added DIGITAL_TWIN_BUILDER_FLOW type to the functions? I see it was added to only several method.. why?

tests/test_commands/test_cp.py
```diff
@@ -18,7 +18,10 @@
 from fabric_cli.errors import ErrorMessages
 from tests.test_commands.commands_parser import CLIExecutor
 from tests.test_commands.conftest import (
```
aviatco, line 20: was something changed here? or only the formating?

tests/test_commands/recordings/test_commands/test_stop/class_setup.yaml
```diff
@@ -11,12 +11,12 @@ interactions:
       Content-Type:
```
aviatco, line 11: recordings for export are missing?

tests/test_commands/test_mkdir.py
```diff
@@ -100,9 +100,7 @@ def test_mkdir_unsupported_item_failure(
         workspace = workspace_factory()
 
         # Create unsupported item
```
aviatco, line 102: also here, only formatting was changed?
try use the same formatting so you'll not have those diffs - will be easier to review

tests/test_commands/test_cp.py
v-alexmoraru: Reverted the auto-formatting

tests/test_commands/test_mkdir.py
v-alexmoraru: Reverted the auto-formatting.

tests/test_commands/recordings/test_commands/test_stop/class_setup.yaml
v-alexmoraru: No, they're included

tests/test_commands/conftest.py
v-alexmoraru: I'd keep this file's formatting as it is, I think it's cleaner.
I added the new item to the supported operations and cases only. It requires an underlying Digital Twin Builder, so it's not acting like any other basic item type

src/fabric_cli/core/fab_config/command_support.yaml
```diff
@@ -161,7 +161,8 @@ commands:
           # - kql_database
           - mirrored_database
           - cosmos_db_database
-          # - digital_twin_builder
+          - digital_twin_builder
+          - digital_twin_builder_flow
```
ayeshurun: @v-alexmoraru i think it was problematic, since digital twin builder creates some additional items implictly and when you move to another location, it was unclear what is the expected behavior with the ones automatically created, should we move them too? hence we left as unsupported.

src/fabric_cli/core/fab_config/command_support.yaml
```diff
@@ -191,7 +192,8 @@ commands:
...
```
ayeshurun: @v-alexmorarui same comment as for `mv`

ayeshurun (dismissed):

ayeshurun (approved):

PR #263, opened 2026-07-10
feat: support enable_shortcut_publish feature flag for deploy command

mprice-gcmlp:
## Summary

By default, fab deploy does not include shortcuts in the deployment of Lakehouses. There is a feature flag in fabric-cicd to allow this though. 

The feature is disabled by default, so existing deployments keep the standard per-item publish behavior. When enabled, the CLI turns on the feature flag being passed to fabric-cicd and will then deploy the shortcuts alongside the Lakehouse.

## Changes
New command flag: Added --shortcut_publish (if not provided, default is false).
Deploy behavior: deploy_with_config_file now calls _apply_shortcut_publish_feature_flags(args), which appends the enable_shortcut_publish  feature flag.

### DIFF
--- a/.changes/unreleased/added-20260710-000000.yaml
+++ b/.changes/unreleased/added-20260710-000000.yaml
@@ -1,0 +1,6 @@
+kind: added
+body: Adds the `--shortcut_publish` flag to the `deploy` command to deploy shortcuts with lakehouses
+time: 2026-07-10T00:00:00.000000+00:00
+custom:
+    Author: mprice-gcmlp
+    AuthorLink: https://github.com/mprice-gcmlp
--- a/docs/commands/fs/deploy.md
+++ b/docs/commands/fs/deploy.md
@@ -69,6 +69,25 @@ Notes:
 
 ---
 
+### Shortcut Publish
+
+By default, `deploy` does **not** deploy shortcuts (defined in `shortcuts.metadata.json`)
+together with their Lakehouse. You can optionally enable **shortcut publish** to include them.
+
+Enable it with the `--shortcut_publish` flag:
+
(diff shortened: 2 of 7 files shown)


mprice-gcmlp:
@microsoft-github-policy-service agree [company="GCM Grosvenor"]

mprice-gcmlp:
@microsoft-github-policy-service agree company="GCM Grosvenor"

ayeshurun:
Thanks for the contribution, @mprice-gcmlp. This is well put together, clean, tested, and documented, and it prompted me to take a proper look at how feature flags flow through deploy.

After digging in, I think we can solve this without any CLI change. The deploy command's config file already has a top-level features field for exactly this, so enable_shortcut_publish (and any other fabric-cicd feature flag) can be enabled directly in config.yml.

```yaml
core:
  workspace_id: "..."
  repository_directory: "."
features:
  - enable_shortcut_publish
```

_References_:

- Deploy configuration file, see the features field: https://microsoft.github.io/fabric-cli/commands/fs/deploy/#configuration-file
- fabric-cicd feature flags: https://microsoft.github.io/fabric-cicd/latest/how_to/optional_feature/#feature-flags

Since the config path already covers this cleanly, adding a dedicated --shortcut_publish flag would grow the CLI surface for something the config already handles, and the same reasoning applies to feature flags in general. So rather than a code change, I would suggest using the features section of the config file.

Given that, I am going to close this PR. If you still think there is a gap that the config approach does not cover, please open an issue as a feature request and we will triage it and discuss the design and the proper solution together.

Really appreciate you raising this and putting the work in. Happy to help if anything about the config approach is unclear.

PR #265, opened 2026-07-15
feat: Add Azure CLI auth source

### LINKED ISSUE #50: [FEATURE] Support Azure Cli credential
### Use Case / Problem

Current auth methods supported by fabric-cli are very limited.

In typical workflows, credentials could be passed along with the context rather than requiring a separate login.

### Proposed Solution

Adding support for Azure CLI credentials as an authentication method would be a significant improvement. 

This would allow fabric-cli to leverage any authentication mechanism supported by Azure CLI and save your effort to dup  the implementation.

### Alternatives Considered

_No response_

### Impact Assessment

- [ ] This would help me personally
- [ ] This would help my team/organization
- [x] This would help the broader fabric-cli community
- [ ] This aligns with Microsoft Fabric roadmap items

### Implementation Attestation

- [x] I understand this feature should maintain backward compatibility with existing commands
- [x] I confirm this feature request does not introduce performance regressions for existing workflows
- [x] I acknowledge that new features must follow fabric-cli's established patterns and conventions

### Implementation Notes

_No response_

shirasassoon:
- Resolves #50

## Summary

Add `azure_cli` as a new identity type that delegates token acquisition to Azure CLI via `azure-identity`'s `AzureCliCredential`. This allows tools calling `fab` to reuse an existing `az login` session instead of requiring a separate interactive `fab auth login`.

## Design Decisions

Following security review, the implementation was simplified:

- **No drift detection** — the user consciously manages their `az login` session. If identity changes, Fabric APIs naturally reject with 401/403.
- **No client-side token cache** — `AzureCliCredential` handles caching internally via Azure CLI's MSAL cache.
- **No `--tenant` flag** for Azure CLI mode (P0) — tenant is derived from the active `az login` session.
- **No rollback on login failure** — consistent with MSAL auth flows.

### Core auth (`src/fabric_cli/core/fab_auth.py`)
- Add `azure_cli` to `AUTH_KEYS` identity type allowlist
- Add `_acquire_token_from_azure_cli()` using singleton `AzureCliCredential`
- One-time tenant extraction from JWT on first token call (stored for context)
- Graceful error handling: `CredentialUnavailableError` → clear "not installed/not logged in" message

### Command handler (`src/fabric_cli/commands/auth/fab_auth.py`)
- Add `--azure-cli` flag to `fab auth login`
- Add "Azure CLI (existing 'az login' session)" option to interactive login menu
- Login flow: `set_access_mode("azure_cli")` → `_acquire_default_access_tokens()` (3 scopes) → set tenant context
- Consistent with MSAL login flow (no success message, no rollback)

### Error handling (`src/fabric_cli/errors/auth.py`)
- `azure_cli_not_available` — Azure CLI not installed or not logged in
- `azure_cli_auth_failed` — authentication failure
- `azure_cli_token_acquisition_failed` — token acquisition failure with scope info

### Documentation
- `docs/commands/auth/index.md` — `--azure-cli` flag in command reference
- `docs/examples/auth_examples.md` — Azure CLI auth examples

### Dependencies
- Add `azure-identity>=1.15.0`

## Tests

**28 automated tests** across 3 files:

| File | Count | Coverage |
|------|-------|----------|
| `test_fab_auth_azure_cli.py` | 24 | Core: token acquisition, singleton, scopes, lifecycle, JWT decode, auth isolation |
| `test_fab_msal_bridge_azure_cli.py` | 2 | Bridge: `MsalTokenCredential` dispatch with azure_cli identity |
| `test_auth.py` (command) | 4 | Command: flag login, tenant ignored, interactive menu, failure propagation |

Key test scenarios:
- Auth method isolation — user/SPN don't invoke `AzureCliCredential`, azure_cli doesn't invoke MSAL
- 3-scope token acquisition at login (Fabric, OneLake, Azure)
- Azure CLI not installed / not logged in error handling
- Singleton credential reuse
- Login/logout lifecycle
- MSAL regression verified (273 core tests pass)

## Security

- Uses `AzureCliCredential` from `azure-identity` SDK — no direct `az` subprocess calls
- Error messages surfaced from SDK (pre-sanitized by azure-identity)
- Auth methods are mutually exclusive — no credential path overlap
- No tokens cached client-side; relies on Azure CLI's own MSAL cache
- Identity changes handled naturally by API-level authorization (401/403)


src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,7 +28,15 @@ def init(args: Namespace) -> Any:
     # Clean up stale context files when logging in
     Context().cleanup_context_files(cleanup_all_stale=True, cleanup_current=False)
 
-    if args.identity:
+    if getattr(args, "azure_cli", False):
+        FabAuth().set_access_mode("azure_cli", args.tenant)
+        FabAuth().set_azure_cli(args.tenant)
```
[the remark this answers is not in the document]
shirasassoon: Confirmed. Because `--azure-cli` is checked first, it silently takes precedence over managed-identity and service-principal options. This is consistent with a pre-existing issue in the authentication argument handling rather than being unique to Azure CLI: managed identity also silently ignores incompatible service-principal credential flags, and service-principal credential methods have implicit precedence. One exception is `-identity -u`, where -u intentionally specifies a user-assigned managed identity client ID. The feedback is valid, and conflict validation should ideally be addressed consistently across all authentication modes.

tests/test_core/test_fab_auth_azure_cli.py
```diff
@@ -0,0 +1,262 @@
...
+@pytest.fixture
+def auth_instance(temp_dir_fixture):
+    """Get a fresh FabAuth instance."""
+    # Clear singleton for test isolation
+    FabAuth.__wrapped__ = None  # type: ignore
+    from fabric_cli.core import fab_auth as fab_auth_module
+
+    if FabAuth in fab_auth_module.singleton.__wrapped__:  # type: ignore
+        del fab_auth_module.singleton.__wrapped__[FabAuth]  # type: ignore
+    return FabAuth()
```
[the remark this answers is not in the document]
shirasassoon: No action needed. The fixture already resets the singleton's state fields between tests, and all 23 tests pass with proper isolation.

src/fabric_cli/core/fab_auth.py
```diff
@@ -418,6 +430,133 @@ def set_managed_identity(self, client_id=None):
...
+        except Exception as e:
+            # Allowlist: SDK exceptions are pre-sanitized by azure-identity; unknown exceptions get a safe generic message
+            if type(e).__name__ in ("ClientAuthenticationError", "HttpResponseError"):
+                error_msg = str(e)
+            else:
+                error_msg = ErrorMessages.Auth.azure_cli_token_acquisition_failed()
+            raise FabricCLIError(
+                ErrorMessages.Auth.azure_cli_auth_failed(error_msg),
+                status_code=con.ERROR_AUTHENTICATION_FAILED,
+            )
```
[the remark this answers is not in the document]
shirasassoon: Not a concern - these are publicly documented exceptions from azure identity that are known to be sanitized.

src/fabric_cli/core/fab_auth.py
```diff
@@ -418,6 +432,142 @@ def set_managed_identity(self, client_id=None):
...
+        switch tenants via 'az login --tenant'.
+        """
+        # Clear credential to force recreation
+        self._azure_cli_credential = None
+
+        # Acquire a probe token to discover identity from JWT claims
+        try:
+            probe_credential = AzureCliCredential()
+            probe_token = probe_credential.get_token(con.SCOPE_FABRIC_DEFAULT[0])
+            claims = self._decode_jwt_token(probe_token.token)
```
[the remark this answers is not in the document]
shirasassoon: The JWKS limitation and the `iss` check are both only relevant in sovereign cloud scenarios, which are not supported by Fabric CLI (all endpoints, scopes, and authority are hardcoded to commercial). When sovereign cloud support is added, both the JWKS URL resolution and the `iss` check would be updated as part of that effort.

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,12 +27,22 @@ def init(args: Namespace) -> Any:
...
+        FabAuth().set_access_mode("azure_cli", args.tenant)
+        try:
+            FabAuth().set_azure_cli(tenant_id=args.tenant)
+        except Exception:
+            FabAuth().logout()
+            raise
+        _acquire_default_access_tokens(FabAuth())
+        Context().context = FabAuth().get_tenant()
+        tenant_id = FabAuth().get_tenant_id() or "unknown"
+        fab_ui.print_grey(f"✓ Authenticated via Azure CLI (tenant: {tenant_id})")
```
tarostok: do we have similar messages for other success auth?

shirasassoon: Removed for consistency with MSAL auth modes.

docs/examples/auth_examples.md
```diff
@@ -25,6 +25,38 @@ fab auth login
...
+
+Reuse an existing Azure CLI session instead of requiring a separate Fabric CLI login. Useful when tools or scripts already have `az login` done (e.g., in development environments or CI/CD pipelines with Azure CLI pre-authenticated).
+
+!!! info "Requires Azure CLI to be installed and logged in (`az login`)"
+
+#### Log in using Azure CLI in interactive mode
+
+```
+fab auth login
+? How would you like to authenticate Fabric CLI? Azure CLI (existing 'az login' session)
```
aviatco, line 38: should this option be one of the options we suggest on interactive mode?

docs/examples/auth_examples.md
```diff
@@ -25,6 +25,38 @@ fab auth login
...
+```
+
+#### Log in using Azure CLI directly from command line
+
+```
+fab auth login --azure-cli
+```
+
+#### Log in using Azure CLI with tenant flag
```
aviatco: I think this is redundant as will never add any value to the user. not sure I understand why we need this flag

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,12 +27,20 @@ def init(args: Namespace) -> Any:
     # Clean up stale context files when logging in
     Context().cleanup_context_files(cleanup_all_stale=True, cleanup_current=False)
 
-    if args.identity:
+    if getattr(args, "azure_cli", False):
+        FabAuth().set_access_mode("azure_cli", args.tenant)
+        try:
+            FabAuth().set_azure_cli(tenant_id=args.tenant)
+        except Exception:
```
aviatco: consider print to the user that he needs to run az login before

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,12 +27,20 @@ def init(args: Namespace) -> Any:
...
+            FabAuth().logout()
```
aviatco: do we need to run FabAuth().logout()? As the user didnt logged in
Or did you do it for clean the azure cred?

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -69,10 +75,18 @@ def init(args: Namespace) -> Any:
         try:
             if selected_auth == "Interactive with a web browser":
                 FabAuth().set_access_mode("user", args.tenant)
-                FabAuth().get_access_token(scope=fab_constant.SCOPE_FABRIC_DEFAULT)
-                FabAuth().get_access_token(scope=fab_constant.SCOPE_ONELAKE_DEFAULT)
-                FabAuth().get_access_token(scope=fab_constant.SCOPE_AZURE_DEFAULT)
+                _acquire_default_access_tokens(FabAuth())
                 Context().context = FabAuth().get_tenant()
+            elif selected_auth.startswith("Azure CLI"):
+                FabAuth().set_access_mode("azure_cli", args.tenant)
```
aviatco: duplicate logic - consider move it to a reuseable method

src/fabric_cli/commands/auth/fab_auth.py
aviatco, line 300: do we want to print the user identity type?

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -275,6 +288,7 @@ def __mask_token(scope):
 
     auth_data = {
         "logged_in": is_logged_in,
+        "auth_source": identity_type,
```
aviatco: where do we use it?

src/fabric_cli/core/fab_auth.py
```diff
@@ -418,6 +438,163 @@ def set_managed_identity(self, client_id=None):
...
+
+            # Fail-closed: reject tokens with missing identity claims
+            if not claims.get("iss") or not claims.get("tid") or not claims.get("oid"):
+                raise FabricCLIError(
+                    ErrorMessages.Auth.azure_cli_token_missing_claims(),
+                    status_code=con.ERROR_AUTHENTICATION_FAILED,
+                )
+
+            # Tenant drift check (most common drift scenario)
+            if stored_tenant and claims["tid"] != stored_tenant:
```
aviatco: is this the case where user pass the tenant flag but az login using another tenant?

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: Removed. It was initially added to be displayed in auth status, but I realized we don't call out auth mode for MSAL flows so I decided to keep it consistent by not having it present.

src/fabric_cli/core/fab_auth.py
shirasassoon: Passing the tenant flag is one case, but regardless of tenant flag. If a user did az login (t1) then fab auth login --azure-cli (the session uses t1), then if the user does az login (t2) (without logging out of fabric cli) and then runs a command (fab ls), this check will catch the change in tenant.

docs/examples/auth_examples.md
shirasassoon: It is an option on interactive mode.

docs/examples/auth_examples.md
shirasassoon: Are you specifically referring to the tenant flag?

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: We do... the error messages are referenced from within `set_azure_cli()`

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: It is a roll back for the scenario when I do fab auth login (interactive user)/valid fab auth login --azure-cli, and then attempt a new login (fab auth login --azure-cli), but for some reason there's a failure with acquiring the probe token (for example if az is logged out in between fabric cli logins), however, at this point auth.json only will contain identity_type: azure_cli and no tid, oid, and iss information which would mess up future drift checks, so we clear the auth.json by doing logout.

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: We haven't done this for MSAL, is this something we want to change?

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: Fixed

docs/examples/auth_examples.md
aviatco: asking cause it looks like the continuation of the sentence... not so clear

docs/examples/auth_examples.md
aviatco: yes

shirasassoon: If yes, I agree it doesn't add much practical value - arguably it is more declarative and a way for the user to validate the identity that is used from azure cli context going into fabric cli during login.

src/fabric_cli/commands/auth/fab_auth.py
aviatco: no,  I think we shouldn't print it... does passing it to print_output_format will not do that?

shirasassoon: as far as information, it is additive and there is no security concern of exposing anything sensitive.

docs/examples/auth_examples.md
shirasassoon: It follows the same style as the other documented auth options.

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,12 +27,17 @@ def init(args: Namespace) -> Any:
     # Clean up stale context files when logging in
     Context().cleanup_context_files(cleanup_all_stale=True, cleanup_current=False)
 
-    if args.identity:
+    if getattr(args, "azure_cli", False):
+        FabAuth().set_access_mode("azure_cli")
```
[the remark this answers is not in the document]
shirasassoon: This is not relevant

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -27,12 +27,15 @@ def init(args: Namespace) -> Any:
...
+        _acquire_default_access_tokens(FabAuth())
+        Context().context = FabAuth().get_tenant()
```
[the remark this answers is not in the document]
shirasassoon: Fixed - additional info added for azure cli auth mode only for better clarity, ignore PR description

src/fabric_cli/core/fab_auth.py
```diff
@@ -10,6 +10,13 @@
 import jwt
 import msal
 import requests
+from azure.core.exceptions import (
+    ClientAuthenticationError,
+    HttpResponseError,
+    ServiceRequestError,
+    ServiceResponseError,
+)
+from azure.identity import AzureCliCredential, CredentialUnavailableError
```
ayeshurun: i believe we can enhance imports by import those only inside `_acquire_token_from_azure_cli` method.

tests/test_core/test_fab_auth_azure_cli.py
```diff
@@ -0,0 +1,521 @@
...
+    iss: str = "https://sts.windows.net/test-tenant/",
+    **extra_claims,
+) -> str:
+    """Create a fake JWT with specified claims (no signature validation needed)."""
+    header = base64.urlsafe_b64encode(b'{"alg":"none"}').rstrip(b"=").decode()
+    claims = {"tid": tid, "oid": oid, "iss": iss, **extra_claims}
+    payload = (
+        base64.urlsafe_b64encode(_json.dumps(claims).encode()).rstrip(b"=").decode()
+    )
+    return f"{header}.{payload}.fakesig"
```
ayeshurun: i feel uncomfortable with this change (also in the other test file). would it be possible to set 
mock_token.token = "test string"

and assert this instead?

shirasassoon: Updated

src/fabric_cli/core/fab_auth.py
```diff
@@ -418,6 +420,72 @@ def set_managed_identity(self, client_id=None):
...
+            # Create singleton credential — inherit Azure CLI context
+            if self._azure_cli_credential is None:
+                self._azure_cli_credential = AzureCliCredential()
+            # AzureCliCredential.get_token expects scopes as positional args
+            azure_token = self._azure_cli_credential.get_token(scope[0])
+
+            claims = self._decode_jwt_token(azure_token.token)
+            tid = claims.get("tid")
+            if tid and tid != self.get_tenant_id():
+                self._synchronize_azure_cli_tenant(tid)
```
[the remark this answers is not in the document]
shirasassoon: Tenant drift is intentionally allowed. Fabric CLI does not pin or reject changes to the active Azure CLI tenant; Azure CLI remains the source of truth. The token tenant comparison is used only to synchronize Fabric CLI’s stored tenant, context, and tenant-scoped caches when the user changes tenants. We will describe this as drift synchronization rather than no drift detection.

We still compare the token’s`tid` because Fabric CLI retains tenant-scoped state, including the persisted tenant, current hierarchy context, and workspace/folder caches. Without detecting the tenant change, Fabric CLI could combine a tenant-B access token with tenant-A context or cached resources, causing stale results and confusing authorization or not-found errors.

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -273,21 +276,80 @@ def __mask_token(scope):
...
+        "certificate": "--certificate",
+        "federated_token": "--federated-token",
+    }
+    specified_args = [
+        attribute for attribute in auth_args if getattr(args, attribute, None)
+    ]
+
+    incompatible: list[str] = []
+    if "azure_cli" in specified_args:
+        incompatible = specified_args if len(specified_args) > 1 else []
```
ayeshurun: let's keep at the moment only validation for azure cli and remove the rest.

docs/commands/auth/index.md
```diff
@@ -32,6 +52,7 @@ fab auth login [-u <client_id>] [-p <client_secret>] [--federated-token <token>]
 - `-p, --password`: Client secret for service principal. Optional.
 - `--federated-token`: Federated token for workload identity. Optional.
 - `--certificate`: Path to certificate file. Optional.
+- `--azure-cli`: Use an existing Azure CLI login session as the token provider. Requires Azure CLI to be installed and logged in (`az login`). Optional.
```
ayeshurun, line 61: lets clarify that --tenant cannot be used with --azure-cli

src/fabric_cli/core/fab_logger.py
```diff
@@ -273,3 +273,12 @@ def _parse_json_into_single_line(json_text, ctxt_cmd):
         return compact_json
     except json.JSONDecodeError:
         return "Failed to parse JSON response"
+
+
+def suppress_azure_sdk_logging() -> None:
```
ayeshurun, line 278: can it being called implictly somehow?

tests/test_commands/test_auth.py
```diff
@@ -903,30 +920,78 @@ def test_auth_status(self, mock_fab_auth, capsys):
...
+        auth.get_access_token.side_effect = FabricCLIError(
+            ErrorMessages.Auth.azure_cli_not_available(),
+            fab_constant.ERROR_AUTHENTICATION_FAILED,
+        )
+
+        with patch.object(auth, "get_identity_type", return_value="azure_cli"):
+            fab_auth.status(args)
+
+        captured = capsys.readouterr()
+        assert "Not logged in to app.fabric.microsoft.com" in captured.err
```
aviatco, line 987: this is new logic that was added? maybe I missed it

tests/test_commands/test_auth.py
```diff
@@ -949,6 +1014,180 @@ def test_init_when_user_cancels_the_prompt(
...
+
+        result = fab_auth.init(args)
+
+        mock_fab_auth_instance = mock_fab_auth.get("instance")
+        mock_fab_auth_instance.set_access_mode.assert_called_with("azure_cli")
+        assert_get_access_token(mock_fab_auth_instance)
+        assert result is True
+
+    @pytest.mark.parametrize(
+        ("other_auth_arg", "expected_flag"),
```
aviatco, line 1028: nit - why not define id here as well?

tests/test_commands/test_auth.py
```diff
@@ -949,6 +1014,180 @@ def test_init_when_user_cancels_the_prompt(
...
+        self, mock_fab_auth, other_auth_arg, expected_flag
+    ):
+        """Azure CLI auth cannot be combined with another auth mode."""
+        args = prepare_auth_args({"azure_cli": True, **other_auth_arg})
+
+        with pytest.raises(FabricCLIError) as ex:
+            fab_auth.init(args)
+
+        assert ex.value.status_code == fab_constant.ERROR_INVALID_INPUT
+        assert "--azure-cli" in ex.value.message
```
aviatco: why not compare the message to be equals to predefine msg?

tests/test_commands/test_auth.py
```diff
@@ -949,6 +1014,180 @@ def test_init_when_user_cancels_the_prompt(
...
+            ),
+        ],
+    )
+    def test_init_with_azure_cli_flag_rejects_other_auth_args(
+        self, mock_fab_auth, other_auth_arg, expected_flag
+    ):
+        """Azure CLI auth cannot be combined with another auth mode."""
+        args = prepare_auth_args({"azure_cli": True, **other_auth_arg})
+
+        with pytest.raises(FabricCLIError) as ex:
```
aviatco, line 1052: what is causing the rejection?

tests/test_commands/test_auth.py
```diff
@@ -949,6 +1014,180 @@ def test_init_when_user_cancels_the_prompt(
...
+        """If token acquisition fails, error propagates (no rollback, consistent with MSAL)."""
+        args = prepare_auth_args({"azure_cli": True})
+        mock_fab_auth_instance = mock_fab_auth.get("instance")
+        mock_fab_auth_instance.get_access_token.side_effect = FabricCLIError(
+            "Azure CLI is not installed or not logged in",
+            "auth_error",
+        )
+
+        with pytest.raises(FabricCLIError, match="not installed or not logged in"):
+            fab_auth.init(args)
```
aviatco, line 1114: consider adding verification for the expected message and error code

tests/test_core/test_fab_auth.py
```diff
@@ -587,7 +587,12 @@ def test_fetch_public_key_success(monkeypatch):
     jwks = {"keys": [fake_jwk]}
 
     # Patch requests.get to return our fake JWKS
-    monkeypatch.setattr(requests, "get", lambda url: fake_response_success(jwks))
```
aviatco, line 590: why changed it here but bellow you kept it?

tests/test_core/test_fab_auth.py
```diff
@@ -629,6 +638,23 @@ def test_fetch_public_key_not_found(monkeypatch):
     assert "Public key not found in JWKS" in str(exc_info.value.message)
 
 
+def test_decode_jwt_token_jwks_timeout(monkeypatch):
```
aviatco, line 641: can you explain?

tests/test_core/test_fab_auth_azure_cli.py
```diff
@@ -0,0 +1,446 @@
...
+    mock_token.token = "test string"
+    mock_token.expires_on = int(time.time()) + 3600
+    mock_credential = MagicMock()
+    mock_credential.get_token.return_value = mock_token
+    mock_class.return_value = mock_credential
+    return mock_credential, mock_token
+
+
+@pytest.fixture(autouse=True)
+def temp_dir_fixture(monkeypatch, tmp_path):
```
aviatco: condiser rename it as this fixture does more then temp_dir

tests/test_core/test_fab_auth_azure_cli.py
```diff
@@ -0,0 +1,446 @@
...
+
+        assert auth.get_tenant_id() is None
+        with patch.object(
+            auth, "_decode_jwt_token", return_value={"tid": "discovered-tenant"}
+        ):
+            auth._acquire_token_from_azure_cli(con.SCOPE_FABRIC_DEFAULT)
+        assert auth.get_tenant_id() == "discovered-tenant"
+
+    @patch("fabric_cli.core.fab_auth.AzureCliCredential")
+    def test_tenant_updated_on_subsequent_calls(
```
aviatco: pls add prefix for _suceess / _failure

tests/test_core/test_fab_auth_azure_cli.py
```diff
@@ -0,0 +1,446 @@
...
+        assert fab_mem_store._get_workspaces_from_cache.cache.currsize == 0
+        assert fab_mem_store._get_workspace_folders_from_cache.cache.currsize == 0
+        mock_credential_class.assert_called_once()
+        assert mock_credential.get_token.call_count == 2
+
+
+class TestAzureCliTokenAcquisition:
+    """Test token acquisition via AzureCliCredential."""
+
+    @patch("fabric_cli.core.fab_auth.AzureCliCredential")
```
aviatco, line 90: consider create it as global fixture for reusing

tests/test_core/test_fab_msal_bridge_azure_cli.py
```diff
@@ -0,0 +1,68 @@
...
+import pytest
+from azure.core.exceptions import ClientAuthenticationError
+
+from fabric_cli.core import fab_constant as con
+from fabric_cli.core.fab_auth import FabAuth
+from fabric_cli.core.fab_msal_bridge import MsalTokenCredential
+
+
+@pytest.fixture(autouse=True)
+def temp_dir_fixture(monkeypatch, tmp_path):
```
aviatco: this is more clear_auth_fixture

tests/test_core/test_fab_msal_bridge_azure_cli.py
```diff
@@ -0,0 +1,68 @@
...
+        "_decode_jwt_token",
+        MagicMock(return_value={"tid": "test-tenant"}),
+    )
+
+
+class TestMsalBridgeAzureCli:
+    """Verify MsalTokenCredential works when identity_type is azure_cli."""
+
+    @patch("fabric_cli.core.fab_auth.AzureCliCredential")
+    def test_bridge_returns_access_token_for_azure_cli(self, mock_credential_class):
```
aviatco: _success*

tests/test_core/test_fab_msal_bridge_azure_cli.py
```diff
@@ -0,0 +1,68 @@
...
+        mock_token.expires_on = int(time.time()) + 3600
+
+        mock_credential = MagicMock()
+        mock_credential.get_token.return_value = mock_token
+        mock_credential_class.return_value = mock_credential
+
+        auth = FabAuth()
+        auth.set_access_mode("azure_cli")
+
+        credential = MsalTokenCredential(auth)
```
aviatco, line 36: why msal? should the test verify that for azure cred the bridge reurn azure cred?

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -16,26 +16,32 @@
...
     utils_mem_store.clear_caches()
 
     # Clean up stale context files when logging in
     Context().cleanup_context_files(cleanup_all_stale=True, cleanup_current=False)
 
-    if args.identity:
+    if getattr(args, "azure_cli", False):
+        FabAuth().set_access_mode("azure_cli")
+        _acquire_default_access_tokens(FabAuth())
+        Context().context = FabAuth().get_tenant()
```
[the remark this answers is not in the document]
shirasassoon: Address this for P1

src/fabric_cli/core/fab_logger.py
shirasassoon: Looked into alternate approach with the help of Copilot. It can be called implicitly when `fab_logger` is imported by invoking `suppress_azure_sdk_logging()` at the end of that module. This applies the policy automatically to every `fab` process, removing the need for an explicit call in `main()`.

docs/commands/auth/index.md
shirasassoon: Fixed

src/fabric_cli/core/fab_auth.py
```diff
@@ -480,6 +551,8 @@ def acquire_token(self, scope: list[str], interactive_renew=True) -> dict:
                         ErrorMessages.Auth.managed_identity_token_failed(),
                         status_code=con.ERROR_AUTHENTICATION_FAILED,
                     )
+            elif identity_type == "azure_cli":
+                token = self._acquire_token_from_azure_cli(scope)
```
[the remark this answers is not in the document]
shirasassoon: Address this for P1

tests/test_commands/test_auth.py
shirasassoon: Yes I added two new fields to fab auth status because of the scenario when az is logged out but fab is still logged in.

shirasassoon: These fields only show when azure cli mode is set.

tests/test_core/test_fab_msal_bridge_azure_cli.py
shirasassoon: Fixed

tests/test_core/test_fab_msal_bridge_azure_cli.py
shirasassoon: The bridge should not return `AzureCliCredential`. The `TokenCredential.get_token()` contract requires an Azure Core `AccessToken`, which is what `MsalTokenCredential.get_token()` returns.

The current test verifies the resulting token, but it could more explicitly verify Azure CLI delegation:

```
mock_credential_class.assert_called_once_with()
mock_credential.get_token.assert_called_once_with(
    con.SCOPE_FABRIC_DEFAULT[0]
)
```
That would prove the bridge used `AzureCliCredential` while retaining the correct assertion that the bridge returns an `AccessToken`. A future cleanup could rename `MsalTokenCredential` to `FabricTokenCredential`, since it now supports MSAL, managed identity, environment tokens, and Azure CLI.

tests/test_core/test_fab_msal_bridge_azure_cli.py
shirasassoon: done

tests/test_core/test_fab_auth_azure_cli.py
shirasassoon: Fixed

tests/test_core/test_fab_auth_azure_cli.py
shirasassoon: Updated

tests/test_core/test_fab_auth_azure_cli.py
shirasassoon: Done

tests/test_core/test_fab_auth.py
shirasassoon: FabAuth._decode_jwt_token() fetches Azure AD’s JWKS public keys when it cannot decode a token with a cached key. This test simulates that JWKS request timing out and verifies the code uses the configured timeout, converts the raw requests.Timeout into a safe authentication error, and does not leak network details.

Safely decoding Azure CLI tokens to discover and synchronize the active tenant. Without bounded JWKS retrieval and sanitized timeout handling, `fab auth login --azure-cli` could hang on network problems or expose low-level connection details; however, this is hardening around the core Azure CLI flow rather than the credential acquisition itself.

tests/test_core/test_fab_auth.py
shirasassoon: It was replaced because production now calls `requests.get(url, timeout=con.AAD_JWKS_TIMEOUT_SECONDS)`, so lambda url: ... would fail with an unexpected timeout argument. The named `fake_get(url, timeout)` is kept because it accepts the current call signature and verifies both the JWKS endpoint and the required bounded timeout.

tests/test_commands/test_auth.py
shirasassoon: Done

tests/test_commands/test_auth.py
shirasassoon: Fixed

tests/test_commands/test_auth.py
shirasassoon: The rejection is caused by `_validate_azure_cli_auth_args(args)` at the start of `fab_auth.init()`. When `azure_cli=True`, it detects any other populated authentication argument, such as `--tenant`, `--identity`, or `--username`, and raises `FabricCLIError` because Azure CLI must use the authentication context from the existing az login session rather than combining credentials from another mode.

tests/test_commands/test_auth.py
shirasassoon: `id` is already defined for each `pytest.param`, for example `id="tenant"`, and `pytest` uses it only to label the generated test case. It should not be added to the parameter tuple because the test body does not consume it; doing so would duplicate metadata as a runtime argument.

src/fabric_cli/core/fab_logger.py
```diff
@@ -273,3 +273,15 @@ def _parse_json_into_single_line(json_text, ctxt_cmd):
...
+    """Prevent Azure SDK records from reaching Fabric CLI output."""
+    for namespace in ("azure.identity", "azure.core"):
+        azure_logger = logging.getLogger(namespace)
+        azure_logger.handlers.clear()
+        azure_logger.addHandler(logging.NullHandler())
+        azure_logger.propagate = False
```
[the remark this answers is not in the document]
shirasassoon: Resolved as “won’t fix for P0.” Fabric CLI is primarily an executable, and import-time Azure SDK suppression is intentional to guarantee clean CLI output; preserving preconfigured handlers could allow SDK logs to leak. We can revisit non-destructive handler behavior later if embedding Fabric CLI as a library becomes a supported scenario.

ayeshurun (approved):

PR #282, opened 2026-08-30
feat(api): add Fabric skill telemetry

shirasassoon:

- Add an internal, global `--skill` argument for command-scoped Fabric skill attribution.
- Store skill attribution in command context so internal requests retain it; non-string values are omitted.
- Add `x-ms-fabric-skill` centrally to Fabric control-plane requests while excluding OneLake, Azure, and Power BI audiences.
- Exclude skill attribution from debug logs, consistent with existing User-Agent telemetry.
- Keep the focused attribution coverage in `test_fab_skill_attribution.py`.

## Validation

- `py -3.13 -m pytest -q tests\test_core\test_fab_skill_attribution.py` (20 passed)
- `py -3.13 -m mypy tests\test_core\test_fab_skill_attribution.py --ignore-missing-imports`
- Targeted Black check passes.


src/fabric_cli/parsers/fab_global_params.py
```diff
@@ -11,11 +13,18 @@ def add_global_flags(parser) -> None:
...
         choices=["json", "text"],
         help="Override output format type. Optional",
     )
+
+    parser.add_argument(
+        "--skill",
+        required=False,
+        default=argparse.SUPPRESS,
+        help=argparse.SUPPRESS,
+    )
```
ayeshurun, line 30: do we need both arg and envar?

src/fabric_cli/core/fab_context.py
```diff
@@ -72,6 +75,18 @@ def command(self) -> str:
...
+    def fabric_skill(self) -> Optional[str]:
+        return self._fabric_skill
+
+    @fabric_skill.setter
+    def fabric_skill(self, fabric_skill: Optional[str]) -> None:
+        if fabric_skill is not None and not re.fullmatch(
+            r"[A-Za-z0-9][A-Za-z0-9._-]{0,127}", fabric_skill
+        ):
+            fabric_skill = None
+        self._fabric_skill = fabric_skill
```
ayeshurun: not sure it is necessary, we can just read the args.skill in do_request instead, making the changes in this PR even smaller. (also no need the fab_decorator change).

src/fabric_cli/client/fab_api_client.py
ayeshurun, line 60:
```suggestion
    fabric_skill = getattr(args, "skill", None)
    json_file = getattr(args, "json_file", None)
```

src/fabric_cli/client/fab_api_client.py
```diff
@@ -132,6 +136,12 @@ def do_request(
                 fab_constant.ERROR_INVALID_OPERATION,
             )
 
+    if audience_value not in ("storage", "azure", "powerbi") and fabric_skill:
```
ayeshurun:
```suggestion
    if fabric_skill and audience_value == "fabric":
```

src/fabric_cli/client/fab_api_client.py
```diff
@@ -132,6 +136,12 @@ def do_request(
...
+        for header_name in list(headers):
+            if header_name.lower() == fab_constant.FABRIC_SKILL_HEADER:
+                del headers[header_name]
+        headers[fab_constant.FABRIC_SKILL_HEADER] = fabric_skill
```
ayeshurun: why is it necessary?

aren't we set the skill header anways?

src/fabric_cli/parsers/fab_global_params.py
shirasassoon: As discussed we will only support arg for P0

src/fabric_cli/core/fab_context.py
shirasassoon: I removed the regex validation, but the context storage and decorator need to be kept ensuring that attribution gets propagated to internal Fabric requests. Because `--skills` exists only on the original command's `args`, internal requests often use newly created `Namespace()` objects without `skill`. The decorator copies the value into shared command context, then every `do_request()` can retrieve it, including path lookups and LRO polling. It also clears the previous value when a later command has no `--skill`, preventing incorrect attribution in interactive sessions.

src/fabric_cli/client/fab_api_client.py
shirasassoon: `fabric_skill = getattr(args, "skill", None)` only works for the initial request. Internal lookups and LRO polling create fresh `Namespace()` objects without `skill`, so attribution would be lost. That's why `fabric_skill` must read from shared command context instead.

src/fabric_cli/client/fab_api_client.py
shirasassoon: Fabric requests can use `audience=None` as well, so we can switch the exclusion logic to:
`if fabric_skill and audience_value in (None, "fabric"):` 
more accurately or leave what we have

src/fabric_cli/client/fab_api_client.py
shirasassoon: We can remove the case normalization logic and just keep as:

`headers[fab_constant.FABRIC_SKILL_HEADER] = fabric_skill`

src/fabric_cli/client/fab_api_client.py
ayeshurun: got it. thanks for the explanation

tests/test_core/test_fab_skill_attribution.py
```diff
@@ -0,0 +1,155 @@
...
+    "command",
+    [
+        ["--skill", "semantic-model-authoring", "export", "ws.Workspace", "-o", "out"],
+        ["export", "--skill", "semantic-model-authoring", "ws.Workspace", "-o", "out"],
+        ["export", "ws.Workspace", "-o", "out", "--skill", "semantic-model-authoring"],
+        ["--skill", "semantic-model-authoring", "job", "run", "ws.Notebook"],
+        ["job", "run", "ws.Notebook", "--skill", "semantic-model-authoring"],
+    ],
+)
+def test_skill_argument_is_preserved_across_command_tree(command):
```
ayeshurun: please add suffix to tests  '_success' and '_failure'

tests/test_core/test_fab_skill_attribution.py
```diff
@@ -0,0 +1,155 @@
...
+        do_request(args)
+
+    request_headers = mock_request.call_args.kwargs["headers"]
+    assert (
+        request_headers[fab_constant.FABRIC_SKILL_HEADER] == "semantic-model-authoring"
+    )
+    assert "X-MS-FABRIC-SKILL" not in request_headers
+
+
+def test_skill_header_is_not_logged(monkeypatch):
```
ayeshurun: please add tests for cases when --skill isn't a string

src/fabric_cli/client/fab_api_client.py
```diff
@@ -113,7 +113,11 @@ def do_request(
     # Build headers
     from fabric_cli.core.fab_context import Context as FabContext
 
-    ctxt_cmd = FabContext().command
+    fab_context = FabContext()
+    ctxt_cmd = fab_context.command
+    fabric_skill = fab_context.fabric_skill
+    if not isinstance(fabric_skill, str):
+        fabric_skill = None
```
ayeshurun: in what scenarios it wont be a string? if feasible, perhaps we should ensure that inside the setter?

tests/test_core/test_fab_skill_attribution.py
shirasassoon: Done

ayeshurun (approved):

PR #283, opened 2026-09-02
feat: Proxy authentication mode with mock token

tarostok:
# 📥 Pull Request

Adds an explicit proxy authentication mode controlled by
`FAB_PROXY_AUTH_ENABLED=true` or `1`.

When enabled, the CLI:

- Uses `mockToken` for Fabric, OneLake, Azure, and custom scopes.
- Overrides configured authentication methods and token environment variables.
- Skips MSAL initialization and JWT validation.
- Supplies expiration metadata for Azure `TokenCredential` compatibility.
- Returns `proxy authentication mode` from `fab auth status`.
- Prevents login and logout with actionable error messages.
- Sends `Bearer mockToken` through the API client for downstream proxy replacement.

This enables environments where an authentication proxy replaces the placeholder
token with the appropriate credential.

No additional dependencies are required.

Tests cover mode detection, authentication precedence, supported scopes, API
headers, JWT bypass, auth commands, and the MSAL credential bridge.

### DIFF
--- a/src/fabric_cli/core/fab_auth.py
+++ b/src/fabric_cli/core/fab_auth.py
@@ -28,6 +28,17 @@ from fabric_cli.core.hiearchy.fab_tenant import Tenant
 from fabric_cli.errors import ErrorMessages
 from fabric_cli.utils import fab_ui as utils_ui
 
+_PROXY_AUTH_ENVIRONMENT_VARIABLE = "FAB_PROXY_AUTH_ENABLED"
+_PROXY_AUTH_PLACEHOLDER_TOKEN = "mockToken"
+_PROXY_AUTH_EXPIRES_ON = 9999999999
+
+
+def _is_proxy_auth_placeholder(token: Any) -> bool:
+    return token in (
+        _PROXY_AUTH_PLACEHOLDER_TOKEN,
+        _PROXY_AUTH_PLACEHOLDER_TOKEN.encode(),
+    )
+
 
 def singleton(class_):
     instances = {}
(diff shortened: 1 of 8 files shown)


tests/test_core/test_fab_auth.py
```diff
@@ -488,6 +496,68 @@ def test_validate_jwt_token_empty():
...
+    monkeypatch.setattr(
+        auth,
+        "_fetch_public_key_from_aad",
+        lambda token: pytest.fail("Proxy authentication must not fetch a public key"),
+    )
+
+    assert auth._decode_jwt_token(token) == {}
+
+
+def test_get_claims_from_token_proxy_auth_mode(monkeypatch):
```
ayeshurun: @copilot please add suffix of '_success' or '_failure' to all tests

src/fabric_cli/core/fab_auth.py
```diff
@@ -525,6 +547,12 @@ def acquire_token(self, scope: list[str], interactive_renew=True) -> dict:
         from fabric_cli.utils.fab_secure_io import restrict_existing_file
 
         try:
+            if self.is_proxy_auth_mode():
```
ayeshurun, line 550: @tarostok do we want/need to validate if env vars tokens exists and if one of env vars tokens exists & proxy mode env var exists, to raise an error?

tarostok: I don't think so, this env var is dedicated to the proxy mode and will override any other auth. Also, it will prevent proxy mode from working in environments with inherited credentials.

ayeshurun (approved):

PR #284, opened 2026-09-03
chore: FIx devcontainer python install issue

ayeshurun:
# 📥 Pull Request

The devcontainer previously relied on the base image's Python setup plus a separate npm-based `changie` install, which was failing dependency installation. This switches to a custom Dockerfile (pinned Python 3.12-bookworm) that installs build deps and `changie` via a verified `.deb` package instead of npm, removes the now-unneeded Node feature, and hardens `install_dev_container_dependencies.sh` (`set -euo pipefail`, absolute paths, optional `PIP_INDEX_URL` override via a git-ignored `local.env`) so dev container setup is reliable across environments, including restricted-network mirrors.

## Changes

- **`.devcontainer/devcontainer.json`**: Switched from a fixed base `image` to `build.dockerfile`, and removed the `node` feature (no longer needed since `changie` is installed via the Dockerfile).
- **`.devcontainer/local.env.example`** (new): Documents an optional `PIP_INDEX_URL` override for use with internal pip mirrors; copy to `.devcontainer/local.env` (git-ignored) to use.
- **`scripts/install_dev_container_dependencies.sh`**: Made executable; added `set -euo pipefail`; resolves paths relative to the repo root; sources `PIP_INDEX_URL` from `.devcontainer/local.env` if present; installs pip requirements only (apt packages and `changie` moved into the Dockerfile).
- **Base image**: Switched from `python:1-3.12-bullseye` (legacy tag format, Debian 11) to `python:3-3.12-bookworm` (current tag format, Debian 12), needed for a valid, maintained base to build the custom Dockerfile from.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,4 +209,5 @@ marimo/_static/
 marimo/_lsp/
 __marimo__/
 
-devcontainer.local.json
+.devcontainer/local.env
+devcontainer-lock.json
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,25 @@
+FROM mcr.microsoft.com/devcontainers/python:3-3.12-bookworm
+
+ARG TARGETARCH
+
+RUN rm -f /etc/apt/sources.list.d/yarn.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        cmake \
+        libcairo2-dev \
+        pkg-config \
+        python3-dev \
+    && changie_version="1.26.0" \
+    && case "${TARGETARCH}" in \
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -209,4 +209,5 @@ marimo/_static/
 marimo/_lsp/
 __marimo__/
 
-devcontainer.local.json
+.devcontainer/local.env
+devcontainer-lock.json
```
aviatco, line 213: shuold be changed from devcontainer.local.json to devcontainer-local.json?

ayeshurun: no. its the file 'local.env' under '.devcontainer' directory.

may-hartov (approved):

PR #285, opened 2026-09-06
refactor: Add user drift detection to azure cli auth mode

shirasassoon:
This pull request introduces robust detection and handling of Azure CLI identity changes in the authentication flow, ensuring that the Fabric CLI logs out and clears cached state if the Azure CLI tenant or principal changes mid-session. It also extends test coverage for these scenarios and improves error messaging for identity drift cases.

**Azure CLI identity drift detection and handling:**

* Added tracking of both tenant and principal IDs (`FAB_TENANT_ID`, `FAB_PRINCIPAL_ID`) in authentication state, and implemented `_check_azure_cli_identity` to detect and respond to identity drift by logging out, clearing caches, and resetting context. [[1]](diffhunk://#diff-51ff78a5157b761c8e6dd8733ba42ed4921d3f4005f29cfa7c6f285dcafb4110L443-R445) [[2]](diffhunk://#diff-51ff78a5157b761c8e6dd8733ba42ed4921d3f4005f29cfa7c6f285dcafb4110L481-R533) [[3]](diffhunk://#diff-7f668d994f2a6aca33cdc0fde940ea9b43f1866673e4291a692b6d14c559e981R57)
* Updated `_acquire_token_from_azure_cli` to validate both tenant and principal IDs from the acquired token, and to call the new drift detection logic. [[1]](diffhunk://#diff-51ff78a5157b761c8e6dd8733ba42ed4921d3f4005f29cfa7c6f285dcafb4110L424-R427) [[2]](diffhunk://#diff-51ff78a5157b761c8e6dd8733ba42ed4921d3f4005f29cfa7c6f285dcafb4110L443-R445)
* Improved error handling and user messaging for missing or changed identity claims, adding new error messages for these scenarios.

**Authentication status and logout flow improvements:**

* Refactored the `status` command to properly handle cases where identity information is missing or has changed, ensuring accurate reporting of login state. [[1]](diffhunk://#diff-3ced7ba1a16925327dd009c58531dfa61b6740843666bc262c1600bc345e9288L223-L224) [[2]](diffhunk://#diff-3ced7ba1a16925327dd009c58531dfa61b6740843666bc262c1600bc345e9288L243-L247) [[3]](diffhunk://#diff-3ced7ba1a16925327dd009c58531dfa61b6740843666bc262c1600bc345e9288R264-R273)
* Ensured that logout and cache clearing are triggered and tested when identity drift is detected.

**Test coverage enhancements:**

* Added comprehensive tests for all Azure CLI identity drift scenarios (tenant, principal, or both changing), for missing claims, and for baseline establishment and updates.
* Updated test fixtures and status command tests to reflect the new identity tracking and error handling. [[1]](diffhunk://#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128L120-R124) [[2]](diffhunk://#diff-ec0bb49e1e306271d6f836cbe691525dd9d392d70fc12ab3ac53cbdfb0046f99L922-R937) [[3]](diffhunk://#diff-ec0bb49e1e306271d6f836cbe691525dd9d392d70fc12ab3ac53cbdfb0046f99R1000-R1048)

These changes make the authentication experience more reliable and secure, preventing accidental use of cached credentials after an Azure CLI identity change.


src/fabric_cli/core/fab_auth.py
```diff
@@ -478,14 +481,62 @@ def _acquire_token_from_azure_cli(self, scope: list[str]) -> dict:
...
+        auth_properties: dict[str, str] = {}
+
+        if current_tenant_id is None:
+            auth_properties[con.FAB_TENANT_ID] = tenant_id
+        if current_principal_id is None:
+            auth_properties[con.FAB_PRINCIPAL_ID] = principal_id
+        if auth_properties:
+            self._set_auth_properties(auth_properties)
+        if current_tenant_id is None:
+            Context().context = self.get_tenant()
```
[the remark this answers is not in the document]
shirasassoon: All Azure CLI login paths call `utils_mem_store.clear_caches()` before changing the authentication mode or acquiring the first token, and command-line processes begin with empty in-memory caches. Explicit logout and detected identity drift also clear caches. Therefore, stale caches cannot survive into initial Azure CLI baseline establishment through the supported CLI flow.

src/fabric_cli/commands/auth/fab_auth.py
```diff
@@ -268,6 +262,18 @@ def __mask_token(scope):
     storage_secret = __mask_token(fab_constant.SCOPE_ONELAKE_DEFAULT)
     azure_secret = __mask_token(fab_constant.SCOPE_AZURE_DEFAULT)
 
+    identity_type = auth.get_identity_type()
+    tenant_id = auth.get_tenant_id()
+
+    if identity_type is None and initial_identity_type == "azure_cli":
```
ayeshurun, line 269: why it is matter if the initial identity was azure_cli? i think if the most recent identity type is None, then we should enter here..

src/fabric_cli/core/fab_auth.py
```diff
@@ -478,14 +485,68 @@ def _acquire_token_from_azure_cli(self, scope: list[str]) -> dict:
...
+            fab_logger.log_warning(f"Change detected in Azure CLI {changed_identity}")
+
+            # Logout, clear caches, and reset context before raising an error
+            self.logout()
+            fab_mem_store.clear_caches()
+            Context().reset_context()
+            raise FabricCLIError(
+                ErrorMessages.Auth.azure_cli_identity_changed(),
+                status_code=con.ERROR_AUTHENTICATION_FAILED,
+            )
```
ayeshurun, line 532: why we raise an error?

src/fabric_cli/core/fab_auth.py
```diff
@@ -478,14 +485,68 @@ def _acquire_token_from_azure_cli(self, scope: list[str]) -> dict:
...
+        auth_properties: dict[str, str] = {}
+
+        if current_tenant_id is None:
+            auth_properties[con.FAB_TENANT_ID] = tenant_id
+        if current_principal_id is None:
+            auth_properties[con.FAB_PRINCIPAL_ID] = principal_id
+        if auth_properties:
+            self._set_auth_properties(auth_properties)
+        if current_tenant_id is None:
+            Context().context = self.get_tenant()
```
ayeshurun, line 543:
```suggestion
        if current_tenant_id is None:
            auth_properties[con.FAB_TENANT_ID] = tenant_id           
        if current_principal_id is None:
            auth_properties[con.FAB_PRINCIPAL_ID] = principal_id
        if auth_properties:
            self._set_auth_properties(auth_properties)
            Context().context = self.get_tenant()
```

src/fabric_cli/core/fab_auth.py
```diff
@@ -478,14 +485,68 @@ def _acquire_token_from_azure_cli(self, scope: list[str]) -> dict:
...
+
+    def _reset_azure_cli_identity_baseline(self):
+        self._auth_info.pop(con.FAB_TENANT_ID, None)
+        self._auth_info.pop(con.FAB_PRINCIPAL_ID, None)
+        self._azure_cli_credential = None
+        self._save_auth()
```
ayeshurun, line 549: do we want to remove all auth_info? consider using self.logout() method. note that if we choose to use self.logout(), it also removes the msal's cache.bin file and resets the config. perhaps we can extract the common code (reset auth file, save auth, set azure_cli_cred to None) to another method and use both, or add a new argument to logout method that tells it to skip remove cache and config..

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: Add a comment

src/fabric_cli/core/fab_decorators.py
```diff
@@ -64,8 +65,14 @@ def decorator(func):
...
-            Context().command = args[0].command_path
-            Context().fabric_skill = getattr(args[0], "skill", None)
+
+            context = Context()
+            context.command = args[0].command_path
+            context.fabric_skill = getattr(args[0], "skill", None)
+            if context.get_runtime_mode() == FAB_MODE_INTERACTIVE:
+                from fabric_cli.core.fab_auth import FabAuth
+
+                FabAuth().validate_azure_cli_identity()
```
ayeshurun, line 75: can you explain this change?

src/fabric_cli/errors/auth.py
```diff
@@ -138,6 +138,17 @@ def azure_cli_token_acquisition_failed() -> str:
             "Run 'az login' to authenticate, then retry"
         )
 
+    @staticmethod
+    def azure_cli_identity_claims_missing() -> str:
+        return "Azure CLI token is missing required tenant or principal identity claims"
```
ayeshurun: for the end user, this error message isn;t actionable or helpful. they cannot control what claims Entra returns in the token to them. We could say generally something like 'Azure CLI token is invalid' or something similar

src/fabric_cli/core/fab_auth.py
shirasassoon: We raise error to prevent the command from completing when identity drift it detected. Without it, even when identity drift is detected, if you do `fab ls` it will return the items in the workspace of the new identity.

src/fabric_cli/errors/auth.py
shirasassoon: Agree. Will update.

src/fabric_cli/core/fab_decorators.py
shirasassoon: Interactive mode keeps authentication, cache, and hierarchy context alive across commands. The user may change their Azure CLI account or tenant in another terminal during that session. This check detects the change before the next command runs, preventing use of stale identity-specific context.

src/fabric_cli/core/fab_auth.py
shirasassoon: This function is used in the following context: when the user explicitly run `fab auth login --azure-cli` again while already authenticated previously with `--azure-cli`. It clears the old tenant/principal baseline and cached `AzureCliCredential`, allowing `_acquire_default_access_tokens()` to establish current `az login` identity as the new baseline instead of throwing an identity drift error. 

When switching from another authentication mode to Azure CLI, the first branch calls the full `logout()` instead so this function isn't called.

src/fabric_cli/commands/auth/fab_auth.py
shirasassoon: Added comment

src/fabric_cli/core/fab_auth.py
shirasassoon: Fixed with your suggestion.

src/fabric_cli/core/fab_auth.py
```diff
@@ -478,14 +485,68 @@ def _acquire_token_from_azure_cli(self, scope: list[str]) -> dict:
...
+            elif tenant_drifted:
+                changed_identity = "Tenant ID"
+            else:
+                changed_identity = "Principal ID"
+            fab_logger.log_warning(f"Change detected in Azure CLI {changed_identity}")
+
+            # Logout, clear caches, and reset context before raising an error
+            self.logout()
+            fab_mem_store.clear_caches()
+            Context().reset_context()
```
ayeshurun, line 527: wondering if it will be better to mode these two calls inside self.logout()

ayeshurun (dismissed):

tests/test_commands/test_auth.py
```diff
@@ -989,6 +997,155 @@ def test_auth_status_azure_cli_session_unavailable(self, mock_fab_auth, capsys):
...
+            output_format="text",
+        )
+        auth = mock_fab_auth["instance"]
+        auth._auth_info = {
+            fab_constant.IDENTITY_TYPE: "azure_cli",
+            fab_constant.FAB_TENANT_ID: "previous-tenant",
+        }
+        auth.get_tenant_id.side_effect = lambda: auth._auth_info.get(
+            fab_constant.FAB_TENANT_ID
+        )
```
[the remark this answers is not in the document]
shirasassoon: In the current fixture, `get_identity_type()` is not replaced with a `MagicMock`; it remains the real method, which reads directly from `auth._auth_info`. It therefore returns "azure_cli" initially and None after the token side effect clears `_auth_info`, so the reacquisition branch is exercised as intended.

`get_tenant_id()` does require the explicit side effect because the fixture mocks it with a fixed return value. For that reason, adding the same configuration for `get_identity_type()` would be redundant, including in the second test.

src/fabric_cli/core/fab_auth.py
shirasassoon: Consider this in the next iteration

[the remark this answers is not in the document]
shirasassoon: Fixed

ayeshurun (approved):

